PR #30, opened 2021-09-21
BugFix/ missing leading zeros when applying hash functions

amine-hmila:
## Summary

When using hash functions like SHA-256, the leading zeros are omitted from the resulting hash.
The generated hash may be used to make comparisons with other systems, this can lead to hash not matching because of the omitted leading zeros

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
Instead of using BigInteger then padding zeros to the result , directly format the bytes array as string 

### How has this been tested?
Unit test 

## Contributor checklist:


hayssams (approved):
LGTM

PR #32, opened 2021-09-30
Field relation/addition in BQ

hayssams:
## Summary
When table is not partitionned and truncate is requested or when table does not exist remove field addition/relaxation options.

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala
```diff
@@ -31,14 +31,33 @@ class BigQuerySparkJob(
...
+            "spark.datasource.bigquery.allowFieldAddition"   -> "false",
+            "spark.datasource.bigquery.allowFieldRelaxation" -> "false"
+          )
+      case Some(_) =>
+        cliConfig.options
+    }
+  }
+
+  override def withExtraSparkConf(): Map[String, String] = {
+    bqOptions
```
seyguai: I think this code is confusing.
`spark.datasource.bigquery.*` is meant to be passed as a spark property, as discussed [here](https://github.com/ebiznext/comet-data-pipeline/issues/582#issuecomment-908987136).
Yet `cliConfig.options` is built from `BigQuerySink#options`, which is not supposed to have any spark config, but only parameters for bigquery source (passed through `DataFrameWriter#option`)
Here, we use the options intended for the bigquery source to provide additional values for a `SparkConfig`.

In the end, `cliConfig.options` still contains `allowField*` when used later in `BigQuerySparkJob#runSparkConnector`.

Furthermore, I read the code used in the bigquery source, to check how those options are handled.
It seems they use the Java Stream API to find the first occurrence of each option in both the `sparkSession#conf` and the provided source options. Which means there is no deterministic order telling us which option will be used if one map contains false and the other true.
Knowing that, I think we need to make sure both the `sparkSession#conf` and `cliConfig#options` have the same `allowField*` values (I think the best way would be to remove them from both map, but we can also force them to false).

seyguai (changes requested):
It lacks the schema update part, since we are disabling entirely this option.
You already did that in here: https://github.com/ebiznext/comet-data-pipeline/pull/599

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala
hayssams: I am wondering if we should not simply assume that the user does not use the spark property by default and when he uses it it means that he wants to force the allowField* semantic.

hayssams: @seyguai Following our conversation
I've differentiated between the connector options and the spark options
I simplified the way we add options to the reader/writer using the options (you were right it is available for some time - since 1.4)

mhdkassir: I didn't get to read all the modifications yet but I think that this PR and previous commits that introduced the method `withExtraSparkConf()`  are trying to handle an issue that should be handled by the user directly. Using allowField* options leads to a BQ error if we are in an Write_Truncate Mode but this is the expected behavior.  I think that KISS principle can fairly be applied here, a wrong configuration options leads to an error and it is up to user to use the right options (whether using spark session or bigquery connector) . Adding code to change the explicit configuration passed by the user adds complexity and can be a source of unexplained behavior (from user point of view) . What do you think ?

hayssams: @mhdkassir We are trying to solve the following issue:
Assume you have a non partitioned table and still is willing to allow field addition/relaxation. This would fail with BigQuery if we do nothing.

This is one of the case we are trying to handle.

mhdkassir: @hayssams perhaps I'm missing something, but if the table is not partitioned and the write mode is write_truncate field addition is possible without adding any options since bigquery will overwrite the schema

hayssams: @mhdkassir When you add data to the table for the first time, you get an error wether you set the allow* option fro true or false.
Below is the message you get from BQ
```Caused by: com.google.cloud.spark.bigquery.repackaged.com.google.cloud.bigquery.BigQueryException: Schema update options should only be specified with WRITE_APPEND disposition, or with WRITE_TRUNCATE disposition on a table partition.```
In that case we need to remove completely the option.

Comment (mhdkassir) [reply src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala]: @hayssams yes but should it be removed by the application code ? I think it is the responsibility of the user to simply correct the configuration by not setting allowField* options at all in such case. I think starlake should not undo an explicit user configuration to mask third party errors. Up to the user to use BigQuery correctly.

Comment (hayssams) [reply src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala]: But then the user need to set different values on first ingestion and for the next ones

Comment (hayssams) [reply src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala]: @mhdkassir When we have a merge operation, the user request an append mode and since he wants field addition/relaxation, he will set those options trop true.
But in fact we end up merge the datasets in memory and do a write_truncate. IF we do no reset those allow* fields, we get the errors above.

Comment (hayssams) [reply src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala]: @seyguai found an interesting bug that would have made us loose the optimisation released with 0.2.6
https://github.com/starlake-ai/starlake/pull/32/files#diff-02d5a7d6430888c78f970e0494a55e832db4a790b8ec2b66c4168a1f012bccfcL908

File: src/main/scala/com/ebiznext/comet/utils/Job.scala [hunk 2]
```diff
@@ -64,10 +65,32 @@ trait JobBase extends StrictLogging with DatasetLogging {
...
+    sourceConfig.remove("spark.datasource.bigquery.allowFieldRelaxation")
+    sourceConfig
+  }
+
+  // Specific to BigQuery
+  val allowFieldAddition: Boolean =
+    if (settings.sparkConfig.hasPath("datasource.bigquery.allowFieldAddition"))
+      settings.sparkConfig.getString("datasource.bigquery.allowFieldAddition").toBoolean
+    else
+      false
```
seyguai: IMO, this should be in `BigQueryJobBase` instead (with an extends on `JobBase`).
And this way you can also check the BQ sink option:
```suggestion
      cliConfig.options.get("allowFieldAddition").exists(_.toBoolean)
```
Same for allowFieldRelaxation

src/main/scala/com/ebiznext/comet/job/index/esload/ESLoadJob.scala
```diff
@@ -42,7 +44,10 @@ class ESLoadJob(
   /** Set extra spark conf here otherwise it will be too late once the spark session is created.
     * @return
     */
-  override def withExtraSparkConf(): Map[String, String] = cliConfig.options
+  override def withExtraSparkConf(sourceConfig: SparkConf): SparkConf = {
+    cliConfig.options.foreach { case (k, v) => sourceConfig.set(k, v) }
```
seyguai: You are using the connector options to fill the spark config.
```suggestion
```

src/main/scala/com/ebiznext/comet/job/index/connectionload/ConnectionLoadJob.scala
```diff
@@ -15,7 +16,10 @@ class ConnectionLoadJob(
   /** Set extra spark conf here otherwise it will be too late once the spark session is created.
     * @return
     */
-  override def withExtraSparkConf(): Map[String, String] = cliConfig.options
+  override def withExtraSparkConf(sourceConfig: SparkConf): SparkConf = {
+    cliConfig.options.foreach { case (k, v) => sourceConfig.set(k, v) }
```
seyguai: You are using the connector options to fill the spark config.
```suggestion
```

src/main/scala/com/ebiznext/comet/utils/Job.scala
```diff
@@ -64,10 +65,32 @@ trait JobBase extends StrictLogging with DatasetLogging {
...
-  def withExtraSparkConf(): Map[String, String] = Map.empty
+  def withExtraSparkConf(sourceConfig: SparkConf): SparkConf = {
+    // During Job execution, schema update are done on the table before data is written
+    // These two options below are thus disabled.
+    // We disable them because even though the suer asked for WRITE_APPEND
+    // On merge, we write in WRITE_TRUNCATE mode.
+    sourceConfig.remove("spark.datasource.bigquery.allowFieldAddition")
+    sourceConfig.remove("spark.datasource.bigquery.allowFieldRelaxation")
+    sourceConfig
+  }
```
seyguai, line 77: IMO this should be in BigQueryJobBase, not here.
This isn't related to Parquet2CSV, or MetricsJob, or ESLoadJob

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala
```diff
@@ -31,14 +31,30 @@ class BigQuerySparkJob(
...
-      cliConfig.options ++ Map(
-        "spark.datasource.bigquery.allowFieldAddition"   -> "false",
-        "spark.datasource.bigquery.allowFieldRelaxation" -> "false"
-      )
-    else
-      cliConfig.options
+  val connectorOptions: Map[String, String] = {
+    val optionsOnTruncate = cliConfig.writeDisposition match {
+      case "WRITE_TRUNCATE" =>
+        cliConfig.options -- List("allowFieldAddition", "allowFieldRelaxation")
```
seyguai: Since now we always update the schema before insertion, we don't really need these options at all.
We can directly remove them, like it is done in `Job#withExtraSparkConf`

src/main/scala/com/ebiznext/comet/utils/Job.scala
hayssams: The Spark session is created in IngestionJob which is a direct subclass of SparkJob. That's why we set it here.
I am also going to handle field addition / relaxation without the user setting any option wince we are governed by the YAML file.

src/main/scala/com/ebiznext/comet/job/index/esload/ESLoadJob.scala
hayssams: Absolutely.
What's wrong ?

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala
hayssams: Removing it in case the user set them

src/main/scala/com/ebiznext/comet/utils/Job.scala
seyguai: But there are some issues with that choice.
- That you have to comment with `Specific to BigQuery` proves that this should not be in the common interface
- You are defining a new default behaviour that will be inherited by all kind of spark jobs, but is highly specific to bigquery.  
- You don't have access to the `cliOptions#options` map, which can also be used by the end-user to configure column update.

src/main/scala/com/ebiznext/comet/job/index/bqload/BigQuerySparkJob.scala
seyguai: Yes, but do we still have to check BQ & writeDisposition anymore ?
Note: this was a suggestion for the whole `connectorOptions` field, but I failed my multi-line selection when doing the review
```suggestion
  val connectorOptions: Map[String, String] = {
    cliConfig.options -- List("allowFieldAddition", "allowFieldRelaxation")
  }
```

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -931,6 +928,56 @@ trait IngestionJob extends SparkJob {
...
+            throw new Exception(
+              s"Merge failed because existing field is nullable but incoming one is required -> $newFieldName"
+            )
+          case Some(false) => // unchanged field
+        }
+      }
+      if (!allowFieldAddition || newFields.isEmpty)
+        updateSchema()
+    }
+  }
```
seyguai: This algorithm does not handle records. You should have a look at MergeUtils#computeCompatibleSchema, it almost does the same thing in the end : if the produced StructType is different from the existing one, then an update is required.
```suggestion
  private def updateBqTableSchema(table: Table, incomingSchema: StructType): Unit = {
    val existingSchema = BigQuerySchemaConverters.toSpark(
      table.getDefinition.asInstanceOf[StandardTableDefinition].getSchema
    )

    val compatibleSchema = MergeUtils.computeCompatibleSchema(existingSchema, incomingSchema)
    if ((allowFieldAddition || allowFieldRelaxation) && compatibleSchema != incomingSchema) {
      val newBqSchema = BigQueryUtils.bqSchema(incomingSchema)
      val updatedTable =
        table.toBuilder.setDefinition(StandardTableDefinition.of(newBqSchema)).build()
      updatedTable.update()
    }
  }
```

src/main/scala/com/ebiznext/comet/utils/Job.scala
hayssams: Actually, it's not a new behaviour.
It's simply that the YAML file is the reference for us. So if the user defined an updated YAML, this means that he wants the field addition / relaxation we are doing.

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
hayssams: Actually we do not need all these checks since BQ will throw an exception if schemas are incompatible

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
```diff
@@ -813,6 +814,7 @@ trait IngestionJob extends SparkJob {
     withScriptFieldsDF: DataFrame,
     mergeOptions: MergeOptions
   ): (DataFrame, Option[List[String]]) = {
+    val incomingSchema = BigQueryUtils.normalizeSchema(schema.finalSparkSchema(schemaHandler))
```
seyguai: You are calling BQ specific code within the parquet path. Is this normal ?
```suggestion
    val incomingSchema = schema.finalSparkSchema(schemaHandler)
```

src/main/scala/com/ebiznext/comet/utils/MergeUtils.scala
```diff
@@ -18,11 +18,15 @@ object MergeUtils extends StrictLogging with DatasetLogging {
...
-      )
+    if (missingColumns.nonEmpty) {
+      logger.info(s"Missing columns ${missingColumns.mkString(",")}")
+      missingColumns.foreach { missingColumn =>
+        if (!actualColumns(missingColumn).nullable)
+          throw new RuntimeException(
+            "Input Dataset should contain every required column from the existing HDFS dataset. The following columns were not matched: " + missingColumn
+          )
+      }
+    }
```
seyguai:
```suggestion
    if (missingColumns.nonEmpty) {
      logger.info(s"Columns omitted in the new schema: ${missingColumns.mkString(",")}")
      val missingColumnsNotNullable = missingColumns.filterNot(actualColumns(_).nullable)
      if (missingColumnsNotNullable.nonEmpty)
        throw new RuntimeException(
          s"Input Dataset should contain every required column from the existing HDFS dataset. The following columns were not matched: ${missingColumnsNotNullable
            .mkString(",")}"
        )
    }
```

src/main/scala/com/ebiznext/comet/job/ingest/IngestionJob.scala
hayssams: You're right!

PR #64, opened 2021-11-25
Get rid of RDD

hayssams:
## Summary
We get rid of the RDD everywhere except in the validators where we manipulate objects of type Any since we are having generic validators here.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/ai/starlake/job/ingest/IngestionJob.scala
+++ b/src/main/scala/ai/starlake/job/ingest/IngestionJob.scala
@@ -999,7 +999,7 @@ object IngestionUtil {
     // We need to save first the application ID
     // refrencing it inside the worker (rdd.map) below would fail.
     val applicationId = session.sparkContext.applicationId
-    val rejectedTypedRDD = rejectedDS.map { err =>
+    val rejectedTypedDS = rejectedDS.map { err =>
       RejectedRecord(
         applicationId,
         now,
@@ -1011,7 +1011,7 @@ object IngestionUtil {
     }
     val rejectedDF = session
       .createDataFrame(
-        rejectedTypedRDD.toDF().rdd,
+        rejectedTypedDS.toDF().rdd,
         StructType(
           rejectedCols.map { case (attrName, _, sparkType) =>
             StructField(attrName, sparkType, nullable = false)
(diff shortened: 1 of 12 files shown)


src/main/scala/ai/starlake/job/ingest/IngestionJob.scala
```diff
@@ -117,33 +91,34 @@ trait IngestionJob extends SparkJob {
...
+      errMessagesDS.take(100).foreach(rejected => logger.debug(rejected.replaceAll("\n", "|")))
     }
     val domainName = domain.name
     val schemaName = schema.name
 
     val start = Timestamp.from(Instant.now())
     val formattedDate = new java.text.SimpleDateFormat("yyyyMMddHHmmss").format(start)
 
-    if (settings.comet.sinkReplayToFile && !rejectedLinesRDD.isEmpty()) {
+    if (settings.comet.sinkReplayToFile && rejectedLinesDS.limit(1).count() > 0) {
```
cchepelov, line 107: a little sad to miss `isEmpty` on datasets. Is this due to back-compatiblity?

src/main/scala/ai/starlake/job/ingest/XmlIngestionJob.scala
```diff
@@ -97,23 +94,23 @@ class XmlIngestionJob(
...
+        saveRejected(rejectedDS, session.emptyDataset[String])
         saveAccepted(
-          dataset,
           ValidationResult(
-            session.sparkContext.emptyRDD[String],
-            session.sparkContext.emptyRDD[String],
-            session.sparkContext.emptyRDD[Row]
+            session.emptyDataset[String],
+            session.emptyDataset[String],
+            dataset
```
cchepelov, line 110: we had `emptyRDD[Row]` now we have `dataset`
Are we sure we're not unintentionally changing behaviour? 
If we're fixing some misbehaving thing, are we sure we now have a unit test to cover this?

cchepelov: oh no, saw the context just before this hunk. Never mind.

src/main/scala/ai/starlake/job/ingest/IngestionJob.scala
hayssams: Exactly.
For a customer you know very well. ```isEmpty``` appeared in Spark 2.4

PR #186, opened 2022-04-05
fix build

ghost:
Fix build as It fails with the recent merge of scala steward PRs.

spark-bigquery-connector 0.24.x doesn't seem to exist
https://github.com/starlake-ai/starlake/runs/5839083148?check_suite_focus=true


hayssams:
Thanks

PR #187, opened 2022-04-05
Feature/enhance extract script

ghost:
## Summary
Add a new attribute to Schema to allow renaming a source from an XLS file. Also renamed the field _rename to _rename_target to avoid confusion.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

_note_: Test will pass after #186 is merged

### Solution
We came across a case where the name of a source table was too long for excel (limited to 31 characters in a tab). So, we needed a way to have the original name in the generated SQL scripts. The idea is to add a new field schema.renameSource  which will be used, if it exists, instead of schema.name in the generated SQL scripts.

### How has this been tested?
Simple unit tests

### Other changes
Renamed schema.rename to schema.renameTarget to avoid confusion.

## Contributor checklist:
- [] My change requires a change to the documentation.


hayssams:
The issue I see here is that it breaks compatibility.
Also I am not aware of this 31 chars limit in Excel cells.
Looking at Excel documentation, lthe limite is above 32000 chars => https://support.microsoft.com/en-us/office/excel-specifications-and-limits-1672b34d-7043-467e-8e27-269d656771c3

mhdkassir:
The limits is on the tab name

https://answers.microsoft.com/en-us/msoffice/forum/all/how-to-increase-the-excel-2016-sheet-name/10750a16-460f-4586-addb-7c43e3875b4e

"This limit is hard coded in Excel programming. This can not be changed by any means."

ghost:
Closed in favor of #190

PR #192, opened 2022-04-07
fix: get a correct sheet name based on the schema name field

ghost:
## Summary
Change type of `private lazy val schemas` from `List[Schema]` to `List[(Schema, SchemaName)]` to use the correct name when getting the sheet name of schema's attributes.

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release notes
 
+# 0.3.14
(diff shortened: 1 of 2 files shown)


ghost:
@hayssams

hayssams (approved):
LGTM

PR #193, opened 2022-04-08
Rename dataset to table in autojob and keep compatibility

hayssams:
## Summary
Rename dataset to table in autojob and keep compatibility. Next step rename schema to table in domain files

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


hayssams:
merged into #194

PR #255, opened 2022-06-28
Fix return value of autojob

ghost:
## Summary
Regression in starlake 0.3.17 where autojobs were always in success even if an error occurs.

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ __New feature__:
 - Support HTTP Stream Source
 - Support XSD Validation
 
+__Bug Fix__:
+- Regression return value of an autojob
+
 # 0.3.17
 __New feature__:
 - Support extra dsv options in conf file
--- a/src/main/scala/com/ebiznext/comet/job/Main.scala
+++ b/src/main/scala/com/ebiznext/comet/job/Main.scala
@@ -106,11 +106,11 @@ object Main extends StrictLogging {
       case "job" | "transform" =>
         TransformConfig.parse(args.drop(1)) match {
           case Some(config) =>
-            if (config.compile)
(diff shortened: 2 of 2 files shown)


ghost:
The build error isnt related to the content of this PR:
```
[error] Failed tests:
[error] 	ai.starlake.schema.handlers.JsonIngestionJobNoIndexNoMetricsNoAuditSpec
[error] 	ai.starlake.schema.handlers.JsonIngestionJobSpecNoIndexJdbcMetricsJdbcAuditSpec
```

hayssams:
Are you sure that we want to keep returning true even when an error happen ?

ghost:
> Are you sure that we want to keep returning true even when an error happen ?

What you describe is the current behavior in master (0.3.17)

```scala
case "job" | "transform" =>
  TransformConfig.parse(args.drop(1)) match {
    case Some(config) =>
      if (config.compile)
        workflow.compileAutoJob(config)
      else
        workflow.autoJob(config)
      true
    case _ =>
      println(TransformConfig.usage())
      false
  }
```

This PR reverts to how it was before, returning true in case of success or false in case of error for an autojob.
```scala
case "job" | "transform" =>
  TransformConfig.parse(args.drop(1)) match {
    case Some(config) =>
      if (config.compile) {
        workflow.compileAutoJob(config)
        true
      } else
        workflow.autoJob(config)
    case _ =>
      println(TransformConfig.usage())
      false
  }
```

I left true to compileAutoJob as Its use is mainly for test/dry-run, right?

hayssams (approved):
LGTM.
Well spotted.
Thanks

PR #385, opened 2023-01-12
fix: allow dataset rename on merge job

tiboun:
## Summary
missing dataset name extraction to deduce projectId from it making jobs such as merge to fail when specifying it.

**Related Issue: #IssueId**
none

**PR Type: Bug Fix **

**Status: Ready to review**

**Breaking change? Yes/No**
No

### Solution
Extract projectId from dataset name first

### How has this been tested?
With unit tests and on our development environment

### Other changes
None

### Deploy Notes
None

## Contributor checklist:
- [X] I have updated the Release notes.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/src/main/scala/ai/starlake/job/sink/bigquery/BigQuerySparkJob.scala
+++ b/src/main/scala/ai/starlake/job/sink/bigquery/BigQuerySparkJob.scala
@@ -295,6 +295,8 @@ class BigQuerySparkJob(
     Utils.logFailure(res, logger)
   }
 
+  override protected def extractProjectDatasetAndTable(resourceId: String): TableId =
+    BigQuerySparkJob.getTableId(session, resourceId)
 }
 
 case class TableMetadata(table: Option[Table], biqueryClient: BigQuery)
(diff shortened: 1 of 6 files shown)


src/main/scala/ai/starlake/job/sink/bigquery/BigQuerySparkJob.scala
```diff
@@ -303,14 +305,21 @@ object BigQuerySparkJob {
...
+    resourceId: String
   ): TableMetadata = {
-    val conf = session.sparkContext.hadoopConfiguration
-    val projectId: String =
-      Option(conf.get("fs.gs.project.id")).getOrElse(ServiceOptions.getDefaultProjectId)
-    val bigquery: BigQuery = BigQueryOptions.getDefaultInstance().getService()
-    val tableId = TableId.of(projectId, datasetName, tableName)
-    TableMetadata(Option(bigquery.getTable(tableId)), bigquery)
+    val finalTableId = getTableId(session, resourceId)
+    val bigquery = BigQueryOptions.getDefaultInstance().getService()
```
hayssams, line 311: This should never be used.
We should call bigquery() that check if a service account has been passed by the user for execution.

This has been added to support SaaS mode where Starlake will serve multiple requests coming from different users.
This is also why we do not have the object BigQueryJobBase anymore and everything has been moved to the trait.

hayssams (approved):

PR #414, opened 2023-02-18
fix: propagate nested fields description

tiboun:
## Summary
Attach field comment to schema for element not on top level.

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**

### Solution
Attach description set from yml schema to spark schema.

### How has this been tested?
Locally with a target on a bq table containing nested fields and arrays.

### Other changes
- Add a provided dependency in order to ease debugging from IDE such as intellij.
- Remove unecessary field from Metadata class

## Contributor checklist:
- [X] I have updated the Release notes.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -121,6 +121,7 @@ object Dependencies {
     // see https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/36
     // Add the jar file to spark dependencies
     "com.google.cloud.spark" %% "spark-bigquery-with-dependencies" % Versions.sparkBigqueryWithDependencies % "provided" excludeAll (jacksonExclusions: _*),
+    "com.google.cloud.spark" % "spark-bigquery-connector-common" % Versions.sparkBigqueryConnectorCommon % "provided" excludeAll (jacksonExclusions: _*),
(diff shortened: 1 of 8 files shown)


project/Dependencies.scala
```diff
@@ -121,6 +121,7 @@ object Dependencies {
     // see https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/36
     // Add the jar file to spark dependencies
     "com.google.cloud.spark" %% "spark-bigquery-with-dependencies" % Versions.sparkBigqueryWithDependencies % "provided" excludeAll (jacksonExclusions: _*),
+    "com.google.cloud.spark" % "spark-bigquery-connector-common" % Versions.sparkBigqueryConnectorCommon % "provided" excludeAll (jacksonExclusions: _*),
```
hayssams, line 124: What is this about ?

hayssams (approved):
LGTM

project/Dependencies.scala
tiboun: This is a dependency used by spark bigquery. Without this, I don't get source code automatically while debugging.

hayssams: Good to know

PR #440, opened 2023-03-08
Feat/add interpolation extract pattern

tiboun:
## Summary
- Turn schema extract pattern to a template
- Add global jdbc schema attributes in order to set common attributes once
- Always propagate domain's metadata to tables
- SMALL BREAKING CHANGE**: Make directory mandatory only when feature require it. If you rely on exception while generating YML files from xls and vice-versa for any missing directory, you'll have to change.

**Related Issue: #IssueId**

**Status: Ready to review**

**Breaking change? Yes**

### How has this been tested?
Add unit tests and use snapshot assembly for client case

## Contributor checklist:

### DIFF
--- a/src/main/scala/ai/starlake/extract/JDBCSchemas.scala
+++ b/src/main/scala/ai/starlake/extract/JDBCSchemas.scala
@@ -4,9 +4,51 @@ import ai.starlake.schema.model.WriteMode
 
 case class JDBCSchemas(
   jdbcSchemas: List[JDBCSchema],
+  globalJdbcSchema: Option[JDBCSchema],
   connectionRef: Option[String] = None,
   connection: Map[String, String] = Map.empty
-)
+) {
+
+  /** @return
+    *   jdbc schemas filled with global jdbc schemas attributes if empty. Considered attributes are
+    *   all except tables:
+    *   - catalog
+    *   - schema
+    *   - tableRemarks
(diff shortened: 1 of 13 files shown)


src/main/scala/ai/starlake/extract/JDBCSchemas.scala
```diff
@@ -4,9 +4,51 @@ import ai.starlake.schema.model.WriteMode
...
+    *   - schema
+    *   - tableRemarks
+    *   - columnRemarks
+    *   - tableTypes
+    *   - template
+    *   - write
+    *   - pattern
+    */
+  def propageGlobalJdbcSchemas(): JDBCSchemas = {
+    if (globalJdbcSchema.isDefined) {
```
hayssams, line 25: Please use "match Some / None" here

src/main/scala/ai/starlake/extract/JDBCSchemas.scala
```diff
@@ -4,9 +4,51 @@ import ai.starlake.schema.model.WriteMode
 
 case class JDBCSchemas(
   jdbcSchemas: List[JDBCSchema],
+  globalJdbcSchema: Option[JDBCSchema],
```
hayssams, line 7: What about defaultJdbcSchema ?

hayssams (approved):
LGTM

PR #448, opened 2023-03-17
WIP: Feature/improve load performance

hayssams:
## Summary
When running the ingestion process, starlake load all domains et schemas into memory. We restrict the loading to the current domain being accessed to improve performance.

**Breaking change? Yes/No**

### Solution
Instead of loading all file sin the folder, we only load the included domains in the command line.


mhdkassir:
👍🏼 I was considering this feature too

mhdkassir:
Can we apply the same concept for jobs ?

hayssams:
Done for domains and merged to master
Still need to be done for jobs

PR #449, opened 2023-03-17
Feat/enhance extract schema

tiboun:
## Summary
Feature :
- Order column extraction from extract-schema according to database column order
- Set domain description if given on bigquery

Fix: 
- use gs if full uri is given even if default fs is file
- remove lowercase on various name in extract-script in order to make it coherent with extract-schema

**PR Type: Bug Fix | Feature**

**Status: Ready to review**

**Breaking change? No**

### How has this been tested?
Unit tests have been modified for column extraction and used in client.
Domain description have been tested in our platform.
The lower case removal fix has been tested in our platform.
The gs have been tested from local to reach gcp via only COMET_ROOT and it works well.
The numeric precision test has been tested in unit tests.

## Contributor checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ __New Feature__:
 - Turn schema extract pattern to a template
 - Add global jdbc schema attributes in order to set common attributes once
 - Always propagate domain's metadata to tables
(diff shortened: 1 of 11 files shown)


hayssams (approved):
LGTM

PR #450, opened 2023-03-18
Use BQ load for simple cases

hayssams:
## Summary
For simple load we use the native bq load functionality for CSV and simple json files
Required AcceptAllValidator and BQ sink to be referenced in the YAML files.

Please note the following limitations:
BQ LOAD CSV/JSON limitations:
- Does not support multi char separator
- Only support YYYY-MM-DD format for date
- Only support YYYY-MM-DD h:mm:ss for timestamp
- Does not support advanced features (field renaming, field deletion, field addition, field transformation) in the YML
- Support BigQuery Schema update options
- Only support newline delimited JSON records
  
**PR Type: Feature**

**Breaking change? No**

### DIFF
--- a/src/main/scala/ai/starlake/extract/JDBCUtils.scala
+++ b/src/main/scala/ai/starlake/extract/JDBCUtils.scala
@@ -178,7 +178,6 @@ object JDBCUtils extends LazyLogging {
       logger.whenInfoEnabled {
         selectedTables.keys.foreach(table => logger.info(s"Selected: $table"))
       }
-
       // Extract the Comet Schema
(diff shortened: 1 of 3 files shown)


elarib:
Hello @hayssams ,
 I believe we can make a simple workaround to make those 3 points possible: 
- Only support YYYY-MM-DD format for date
- Only support YYYY-MM-DD h:mm:ss for timestamp
- Does not support advanced features (field renaming, field deletion, field addition, field transformation) in the YML

The idea is to create a temporary table, and append its data, with applying the right cast, transformation into the final table .

WDYT ?

I can work on it.

hayssams:
Gr8 
go ahead

PR #462, opened 2023-03-30
Feat/keep original domain

tiboun:
## Summary
FEAT

- **BREAKING CHANGE** Add `normalized_domain` variable for schema extraction. `domain` keep the original name.
- keep user changes if set in domain's metadata or table information. Domain-template and other rules that apply to it still have higher precedence.

FIX
- **BREAKING CHANGE** domain template in extract-schema is no more interpolated since we plan to use domains accross multiple environment

### DIFF
--- a/src/main/scala/ai/starlake/extract/ExtractSchema.scala
+++ b/src/main/scala/ai/starlake/extract/ExtractSchema.scala
@@ -80,17 +80,13 @@ object ExtractSchema extends Extract with LazyLogging {
     settings: Settings
   ): Unit = {
 
-    // End of list element have higher precedence
-    def mergeAllMetadata(metadatas: List[Metadata]): Metadata = {
-      metadatas.foldLeft(Metadata())(_.`import`(_))
-    }
-
     val domainName = jdbcSchema.schema.replaceAll("[^\\p{Alnum}]", "_")
     baseOutputDir.createDirectories()
     File(baseOutputDir, domainName).createDirectories()
     val extractedDomain = extractDomain(jdbcSchema, connectionOptions, domainTemplate)
     val domain = extractedDomain.copy(metadata =
-      mergeAllMetadata(Nil ++ currentDomain.flatMap(_.metadata) ++ extractedDomain.metadata)
+      Metadata
+        .mergeAll(Nil ++ currentDomain.flatMap(_.metadata) ++ extractedDomain.metadata)
         .copy(fillWithDefaultValue = false)
         .asOption()
     )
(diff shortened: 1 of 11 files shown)


src/main/scala/ai/starlake/extract/ExtractSchema.scala
```diff
@@ -60,39 +59,82 @@ object ExtractSchema extends Extract with LazyLogging {
...
-        case None => table
-        case Some(write) =>
-          val metadata =
-            domainTemplate.flatMap(_.metadata).getOrElse(Metadata()).copy(write = jdbcSchema.write)
-          table.copy(metadata = Some(metadata))
-      }
+      val restoredTable =
+        currentDomain.flatMap(_.tables.find(_.name == table.name)) match {
+          case Some(currentTable) =>
+            table.copy(
```
hayssams: I think that we should put this code in the Schéma class to centralize access to the fields there

src/main/scala/ai/starlake/schema/handlers/SchemaHandler.scala
```diff
@@ -438,18 +448,22 @@ class SchemaHandler(storage: StorageHandler, cliEnv: Map[String, String] = Map.e
             .map { tableRefName =>
               val schemaPath = new Path(folder, tableRefName)
               YamlSerializer.deserializeSchemaRefs(
-                Utils
-                  .parseJinja(storage.read(schemaPath), activeEnv())
-                  .richFormat(activeEnv(), Map.empty),
+                if (raw)
```
hayssams: Should’nt this be if(!raw)

src/main/scala/ai/starlake/schema/model/Metadata.scala
```diff
@@ -228,6 +247,57 @@ case class Metadata(
     )
   }
 
+  /** Keep metadata that are different only
+    *
+    * @param parent
+    *   : Schema level metadata
+    * @return
+    *   the metadata that differs between parent and this element.
+    */
```
hayssams, line 256: Move this code to Metadata

hayssams (approved):
Could you please make the changes requested 
Thanks for this PR

src/main/scala/ai/starlake/schema/model/Metadata.scala
tiboun: It's already in Metadata, do you think we should move this to the object Metadata ?

hayssams: My bad

PR #465, opened 2023-03-30
feat: trim on numeric for schema extraction

tiboun:
Apply trim on all numeric during schema extraction. Have higher precedence than the one defined in the template.

### DIFF
--- a/src/main/scala/ai/starlake/extract/JDBCUtils.scala
+++ b/src/main/scala/ai/starlake/extract/JDBCUtils.scala
@@ -286,11 +286,9 @@ object JDBCUtils extends LazyLogging {
                 .getOrElse(Nil)
             val selectedColumns =
               if (currentTableRequestedColumns.isEmpty)
-                columns.toList
+                columns
               else
-                columns.toList.filter(col =>
-                  currentTableRequestedColumns.contains(col.name.toUpperCase())
-                )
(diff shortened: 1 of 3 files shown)


src/main/scala/ai/starlake/extract/JDBCUtils.scala
```diff
@@ -287,11 +287,9 @@ object JDBCUtils extends LazyLogging {
                 .getOrElse(Nil)
             val selectedColumns =
               if (currentTableRequestedColumns.isEmpty)
-                columns.toList
+                columns
               else
-                columns.toList.filter(col =>
-                  currentTableRequestedColumns.contains(col.name.toUpperCase())
-                )
+                columns.filter(col => currentTableRequestedColumns.contains(col.name.toUpperCase()))
```
hayssams, line 292: Are we sure that currentTableRequestdeColumns are in upper case ?

hayssams (approved):

src/main/scala/ai/starlake/extract/JDBCUtils.scala
tiboun: yes, it's from the line above:

```
            val currentTableRequestedColumns =
              jdbcTableMap
                .get(tableName)
                .map(_.columns.map(_.toUpperCase))
                .getOrElse(Nil)
```

It is the existing behavior. The change just apply to remove the unecessary `toList` call.

PR #480, opened 2023-04-12
Fix/homogeneous domain folders

tiboun:
## Summary
Improve traceability by using simple domain and table name in areas during ingestion job


hayssams (approved):
Will update the accepted path with finalName

PR #485, opened 2023-04-17
Feat/add more options to extract data

tiboun:
__Feature__:
- allow full export for tables and use partition column only to speed up extraction
- allow to force full export. Useful for re-init cases.

__Bug Fix__:
- Use audit connection settings while fetching last export and its column quotes
- Division by zero when computing progress bar
- Human Readable throw exception when elapsed time is 0
- **Breaking Change** Make table's fetchSize to have higher precedence than the one defined in jdbcSchema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 
 # Release notes
 
+# 0.7.2:
+__Feature__:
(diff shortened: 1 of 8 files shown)


hayssams (approved):
LGTM

PR #487, opened 2023-04-19
Fix/yml2dag list

tiboun:
__Feature__:
- allow to change date and timestamp format during data extraction

__Bug fix__:
- list command for internal ressources doesn't work for yml2dag
- **Breaking Change** Make default configuration of CSV writer in extract-data to match default value of Metadata.
- keep original args in starlake.sh  when they have spaces

### DIFF
--- a/src/main/scala/ai/starlake/extract/SLResultSetHelperService.scala
+++ b/src/main/scala/ai/starlake/extract/SLResultSetHelperService.scala
@@ -1,0 +1,116 @@
+package ai.starlake.extract
+
+/** An adapted code of au.com.bytecode.opencsv.ResultSetHelperService since we can't customize value
+  * by type serialization.
+  */
+
+import au.com.bytecode.opencsv.ResultSetHelper
+
+import java.io.BufferedReader
+import java.sql.{Array => _, _}
+import java.text.SimpleDateFormat
+import java.util
+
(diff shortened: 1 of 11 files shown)


src/main/scala/ai/starlake/extract/SLResultSetHelperService.scala
```diff
@@ -0,0 +1,144 @@
...
+        value = handleTimestamp(rs.getTimestamp(colIndex))
+
+      case Types.NVARCHAR | Types.NCHAR | Types.LONGNVARCHAR =>
+        value = rs.getNString(colIndex)
+
+      case Types.LONGVARCHAR | Types.VARCHAR | Types.CHAR =>
+        value = rs.getString(colIndex)
+
+      case _ =>
+        value = ""
```
hayssams: Should we replace this line with  value = handleObject(rs.getObject(colIndex))

hayssams (changes requested):

src/main/scala/ai/starlake/extract/SLResultSetHelperService.scala
tiboun: Done according to what you suggested :)

hayssams (approved):
LGTM

PR #488, opened 2023-04-20
feat: allow to filter table data before sinking it.

tiboun:
## Summary
allow to filter table data before sinking it.


hayssams (approved):
LGTM

PR #489, opened 2023-04-21
Feat/data extraction filtering

tiboun:
## Summary
- allow to prevent extraction if last extraction is recent enough
- allow to clean all files of table only when it is extracted
- allow to in/exclude schemas and/or tables from extraction

### DIFF
--- a/src/main/scala/ai/starlake/extract/JDBCUtils.scala
+++ b/src/main/scala/ai/starlake/extract/JDBCUtils.scala
@@ -586,7 +586,8 @@ object JDBCUtils extends LazyLogging {
     parallelism: Int,
     fullExportCli: Boolean,
     datePattern: String,
-    timestampPattern: String
+    timestampPattern: String,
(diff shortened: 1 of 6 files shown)


hayssams:
The CLI options are getting much richer 
Could you please add a few sample calls in the extractdataconfig printout ?

src/main/scala/ai/starlake/extract/JDBCUtils.scala
```diff
@@ -606,345 +614,407 @@ object JDBCUtils extends LazyLogging {
...
-
-        // Get the current table partition column and  connection options if any
-        val currentTable = jdbcSchema.tables
-          .flatMap { table =>
-            if (table.name.contains("*") || table.name.equalsIgnoreCase(tableName)) {
-              Some(table)
-            } else {
-              None
-            }
+    val filteredJdbcSchema = if (includeTables.nonEmpty) {
```
hayssams: Should'nt this code be case insensitive for table / schema names ?

PR #517, opened 2023-05-24
Always add domain to dataset area

### LINKED ISSUE #516: [BUG] - Databricks absolute path must be written differently
## Description
I try to distribute my ingestion areas on several containers in an azure data lake mounted on a dbfs path on the databricks platform (ex: /mnt/starlake).
For that I specified in environment variable `SL_AREA_PENDING=dbfs://mnt/pending_area` for example for the pending area.
Problem, when you specify an absolute path on databricks with the filesystem included in it, you have to write it with a single '/'. 

`%fs ls dbfs://mnt/starlake` gives us this error:
**IllegalArgumentException: Hostname not allowed in dbfs uri. Please use 'dbfs:/' instead of 'dbfs://' in uri: dbfs://mnt/starlake**

While `%fs ls dbfs:/mnt/starlake` works correctly.

However I can't change the absolute path of `SL_AREA_PENDING=dbfs:/mnt/pending`_area because according to your code, in **DatasetArea.scala line 49**, a path is recognized as absolute only if it contains "://". 
So if I include the variable this way, starlake will recognize this path as relative and concatenate me SL_DATASETS with SL_AREA_PENDING : **"dbfs:/mnt/datasets_area/dbfs:/mnt/pending_area"**
This will cause an error at runtime.

So it might be useful to change this condition to contains(":/") or just create a specific condition for databricks.

ghost:
Summary
A clear and concise description of the overall goals of the pull request's commits.

**Related Issue: #516 **

**PR Type: Feature **

**Status: Ready to review**

**Breaking change? Yes**

### Solution
Add domain name to dataset area path for absolute dataset area reference
### How has this been tested?
Manually tested on Databricks

### Other changes
Typo fixes

### Remaining Todos
* [ ] - Unit test

## Contributor checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 
 # Release notes
+# 0.7.4:
+- **BREAKING CHANGE**: Add domain name to dataset area path for absolute dataset area reference
 
 # 0.7.3:
 
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -64,7 +64,7 @@ object Common {
 //    }
   )
 
-  def assemlySettings = Seq(
+  def assemblySettings = Seq(
     assembly / test := {},
     Compile / mainClass := Some("ai.starlake.job.Main")
   )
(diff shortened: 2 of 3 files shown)


hayssams (approved):
Gr8 PR
Thanks.
LGTM

PR #552, opened 2023-06-23
feature: improve extraction-schema

tiboun:
extract-schema keep original scripted fields and merge attributes' parameters extract-schema quote catalog, schema and table name.

bugfix: use default privacy level instead of Null.


hayssams:
@tiboun Not sure I understand this PR
I probably need your help on monday.
Especially that the xsd handling seems to have been moved.

tiboun:
@hayssams okay Hayssam, np

hayssams (approved):

PR #557, opened 2023-06-28
Handle automatically merge instructions in transforms

hayssams:
## Summary
When using the merge attribute at the task level, merge are handled automatically by starlake.
**Related Issue: #IssueId**

**Breaking change? Yes/No**

### Solution
Describe your changes in detail

### How has this been tested?
Outline the steps taken to test the PR (unit tests? Integration tests? tested impacts on other project features? ...)

### Other changes
Describe any other additional work done:
bug fixes, tweaks, small refactors

### Deploy Notes
Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, new environment variables, ...

### Remaining Todos
* [ ] - Task1
* [ ] - Task2

## Contributor checklist:
- [ ] My code follows the code style of this project.


src/main/scala/ai/starlake/job/sink/bigquery/BigQueryJobBase.scala
```diff
@@ -374,6 +374,17 @@ trait BigQueryJobBase extends StrictLogging {
     processWithRetry(bigqueryProcess = bigqueryProcess)
   }
 
+  def tableExists(datasetName: String, tableName: String)(implicit settings: Settings): Boolean = {
```
tiboun: we may split getOrCreateTable into two function `getTable` and `createTable`, define getOrCreateTable with those two and use the getTable in order to define this method, what do you think ?

src/main/scala/ai/starlake/job/transform/AutoTask.scala
```diff
@@ -61,8 +61,11 @@ object AutoTask extends StrictLogging {
     settings: Settings,
     schemaHandler: SchemaHandler
   ): Option[String] = {
-    taskDesc.sink
-      .flatMap(_.database) // database defined in sink
+    taskDesc.database
```
tiboun, line 64: In `BigQueryFreshnessInfo`, `AutoTask.createBigQueryConfig`, `resolveTableRefInDomainsAndJobs`, `ingestionWorkflow`,  we may use this getDatabase in order to have only one way of resolving it, what do you think?

Furthermore, why do we have a database attribute in the task and another one in the sink ? Maybe we should keep one of them only.

src/main/scala/ai/starlake/job/transform/AutoTask.scala
```diff
@@ -184,6 +181,14 @@ case class AutoTask(
       new BigQueryNativeJob(config, finalSql, udf)
     }
 
+    val start = Timestamp.from(Instant.now())
+    logger.info(s"running BQ Query  start time $start")
+    val tableExists = bqNativeJob("ignore sql").tableExists(taskDesc.domain, taskDesc.table)
```
tiboun, line 182: Do we move getOrCreateTable/datasets and tableExists to the companion object and have the ability to use them outside of it without instantiating a fake job ?

src/main/scala/ai/starlake/utils/SQLUtils.scala
```diff
@@ -50,6 +56,81 @@ object SQLUtils {
     ctes.map(_.replaceAll("`", ""))
   }
 
+  def extractColumnNames(sql: String): List[String] = {
```
tiboun, line 59: Does jSqlParser infer * select or not ?

```
with
  my_table as (select col1, col2, col3),
  my_table2 as (select col1, col4)
  select my_table.* except(col2), myTable2.col4 from my_table inner join my_table2 using(col1)
```

expected output is col1, col3, col4.

src/main/scala/ai/starlake/utils/SQLUtils.scala
```diff
@@ -50,6 +56,81 @@ object SQLUtils {
...
+        val fullTableName = database match {
+          case Some(db) => s"$quote$db$quote.$quote$domain$quote.$quote$table$quote"
+          case None     => s"$quote$domain$quote.$quote$table$quote"
+        }
+        s"""MERGE INTO
+           |$fullTableName as SL_INTERNAL_SINK
+           |USING($sql) as SL_INTERNAL_SOURCE ON $joinCondition
+           |${buildWhenMatchedUpdate(sql)}
+           |
+           |${buildWhenNotMatchedInsert(sql)}
```
tiboun, line 127: We've got the ability to define `when not matched by source` as well. This allows, for instance, to delete entries in target table.

src/main/scala/ai/starlake/utils/SQLUtils.scala
```diff
@@ -50,6 +56,81 @@ object SQLUtils {
...
+          .mkString(" AND ")
+
+        val fullTableName = database match {
+          case Some(db) => s"$quote$db$quote.$quote$domain$quote.$quote$table$quote"
+          case None     => s"$quote$domain$quote.$quote$table$quote"
+        }
+        s"""MERGE INTO
+           |$fullTableName as SL_INTERNAL_SINK
+           |USING($sql) as SL_INTERNAL_SOURCE ON $joinCondition
+           |${buildWhenMatchedUpdate(sql)}
```
tiboun, line 125: We are forcing users to follow this way of merge request. Do we plan to give the user the ability to override the different merge with templates ? I think we must provide this.

src/main/scala/ai/starlake/utils/SQLUtils.scala
```diff
@@ -50,6 +56,81 @@ object SQLUtils {
...
+           |
+           |${buildWhenNotMatchedInsert(sql)}
+           |
```
tiboun, line 128: merge clauses are not limited, we may have many WHEN MATCHED clause.
The only restriction to merge statement is to have one source line matching at MOST 1 merge clause if I remember well. That is why we can add other check along with the MATCHED clause. Example [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement).

tiboun (changes requested):
Great PR, the beginning of a new user experience is coming in auto jobs

src/main/scala/ai/starlake/job/transform/AutoTask.scala
hayssams: We need to pass it the cliConfig. That's why it's not an object.

src/main/scala/ai/starlake/job/transform/AutoTask.scala
hayssams: About the database field, I agree with you, I guess I need to discuss this one.

hayssams: They all rely on the same function, I you look at the code.

src/main/scala/ai/starlake/job/sink/bigquery/BigQueryJobBase.scala
hayssams: It make sens but I am not able to come up with a clean split.

hayssams: Done

src/main/scala/ai/starlake/job/transform/AutoTask.scala
hayssams: We worked on it together and I cleaned it up

PR #558, opened 2023-06-29
improve starlake output accuracy

tiboun:
__Fix__:
- log failure in audit during ingestion job when unexpected behavior occurs
- wait job output for audit log and RLS queries to avoid any async exception and improve job output result accuracy
- escape string parameters while using native query

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ __Deprecated__:
 - __Bug Fix__:
 - Take into account extensions in domain / metadata attribute
 
+__Fix__:
+- log failure in audit during ingestion job when unexpected behavior occurs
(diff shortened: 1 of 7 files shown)


hayssams (approved):
Didn't know that we could wait for a batch job to finish.
We simply must make sure that this does not slow down too much the ingestion process

tiboun:
> Didn't know that we could wait for a batch job to finish. We simply must make sure that this does not slow down too much the ingestion process

It may impact ingestion job performance as you mentionned but without this wait we could not be sure about the whole ingestion job result. Would you like to have a settings to adjust this behavior and have it disabled by default ?

hayssams:
Yes I think that we better have a flag.

PR #584, opened 2023-08-18
feat: parallel schema extraction

tiboun:
## Summary
schema extraction and features relying on it benefit from parallel fetching


hayssams:
@tiboun  I will merge this and make the following changes:
move the new utils functions to a extractutils class
fix the scalafmt issues

PR #602, opened 2023-09-11
ingest csv unordered and unbalanced columns in native bq mode

tiboun:
ingest csv unordered and unbalanced columns in native bq mode

### DIFF
--- a/src/main/scala/ai/starlake/schema/handlers/HdfsStorageHandler.scala
+++ b/src/main/scala/ai/starlake/schema/handlers/HdfsStorageHandler.scala
@@ -35,9 +35,9 @@ import scala.concurrent.duration.FiniteDuration
 import scala.util.{Failure, Success, Try, Using}
 
 /** HDFS Filesystem Handler
- */
+  */
 class HdfsStorageHandler(fileSystem: String)(implicit
-                                             settings: Settings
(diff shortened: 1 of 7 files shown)


src/main/scala/ai/starlake/schema/handlers/HdfsStorageHandler.scala
```diff
@@ -243,40 +243,55 @@ class HdfsStorageHandler(fileSystem: String)(implicit
...
-    * @return
-    *   file content as a string
-    */
+   *
+   * @param path
+   * : Absolute file path
+   * @return
+   * file content as a string
+   */
   def read(path: Path, charset: Charset = StandardCharsets.UTF_8): String = {
```
hayssams, line 252: We load the whole file in memory right ?

hayssams (approved):
The only limitation I see is the file size.
We'll have to explain it in the docs

src/main/scala/ai/starlake/schema/handlers/HdfsStorageHandler.scala
tiboun: Just refactored the code and added readAndExecute.

Of course I tried not to read the whole file before getting the headers so I think there is no such limit.

PR #618, opened 2023-09-22
enhance dag generation

tiboun:
- add more attributes to generation context
- fix dag generation
- sort collections in generation context

### DIFF
--- a/src/main/scala/ai/starlake/utils/Utils.scala
+++ b/src/main/scala/ai/starlake/utils/Utils.scala
@@ -28,7 +28,8 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.fasterxml.jackson.annotation.{JsonSetter, Nulls}
 import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
-import com.hubspot.jinjava.Jinjava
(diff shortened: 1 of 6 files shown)


src/main/scala/ai/starlake/utils/Utils.scala
```diff
@@ -278,9 +279,17 @@ object Utils extends StrictLogging {
...
 
-  def parseJinjaTpl(templateContent: String, params: Map[String, Object]): String = {
-    val jinjava = new Jinjava()
-    jinjava.render(templateContent, params.asJava);
+  def parseJinjaTpl(templateContent: String, params: Map[String, Object])(implicit
+    settings: Settings
+  ): String = {
+    val config = JinjavaConfig
+      .newBuilder()
+      .withNestedInterpretationEnabled(false)
```
hayssams, line 287: What's the meaning of this ?
Are we relying on it ?

hayssams (approved):
LGTM

src/main/scala/ai/starlake/utils/Utils.scala
tiboun: If you have {{"{{ ds }}"}} then by default it will try to resolve ds var instead of just rendering it as "{{ ds }}". Therefore it would fail.

hayssams: Ok thanks 
Got it 👍

PR #619, opened 2023-09-26
- minimize data to fetch from bigquery

tiboun:
getQueryResults fetch all data. Add query page result options in order to minimize fetched data and avoid out of memory with cloud run jobs, improve performance.


hayssams (approved):
LGTM
We limit the number of rows in interactive mode

PR #621, opened 2023-09-27
fix: remove null values

tiboun:


hayssams:
How come the partition name can be null ?

PR #628, opened 2023-10-03
Fix/bq data fetching

tiboun:
fix data fetching
count null partition rows as rejected record in dynamic partition overwrite

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - **BREAKING CHANGE**
   - STAGE no more used in MetricsJob and Expectations. Existing Metrics Database need to be updated.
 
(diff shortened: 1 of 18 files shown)


hayssams:
I would like to understand this use case. 
Will call you to talk about it

tiboun:
It was related to null partitions in dynamic partition overwrite. When null, we only ignore them. From now on, we count as rejected rows making audit log more accurate since we don't find them in the target table.

This may happen when we use dynamic partition overwrite with a scripted field. Having this dynamic input make it possible to have null values inside and have input record considered as validated.

hayssams:
Ok got it
The scripted field is the partition field in that case !
That's a tough one 👏

PR #707, opened 2023-12-15
to fix proxy settings parser when the character @ is present in credentials

fupelaqu:
## Summary
to fix proxy settings parser when the character @ is present in credentials

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

PR #712, opened 2023-12-20
add dags configuration parameter to generate dags for tasks

fupelaqu:
## Summary
add dags configuration parameter to generate dags for tasks

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

PR #714, opened 2023-12-21
generate all related transforms in one dag

fupelaqu:
## Summary
generate all related transforms in one dag

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/ai/starlake/schema/generator/DagGenerateCommand.scala
+++ b/src/main/scala/ai/starlake/schema/generator/DagGenerateCommand.scala
@@ -110,26 +110,22 @@ class DagGenerateCommand(schemaHandler: SchemaHandler) extends LazyLogging {
       DagPair(k, v)
     }.toList
     val depsEngine = new AutoTaskDependencies(settings, schemaHandler, settings.storageHandler())
-    val taskConfigsWithFilename =
-      taskConfigs
-        .map { taskConfig =>
-          val filename = Utils.parseJinja(
-            taskConfig.dagConfig.filename,
-            schemaHandler.activeEnvVars() ++ Map(
-              "table"  -> taskConfig.taskDesc.table,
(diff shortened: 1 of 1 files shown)


hayssams (approved):
LGTM

src/main/scala/ai/starlake/schema/generator/DagGenerateCommand.scala
```diff
@@ -110,31 +110,40 @@ class DagGenerateCommand(schemaHandler: SchemaHandler) extends LazyLogging {
...
-      applyJ2AndSave(outputDir, jEnv, dagTemplateContent, context.asMap, filename)
-    }
+        (filename, taskConfig)
+      }
+      .groupBy { case (filename, _) => filename }
+    taskConfigsGroupedByFilename
+      .foreach { case (filename, taskConfigs) =>
+        val headConfig = taskConfigs.head match { case (_, config) => config }
+        val dagConfig = headConfig.dagConfig
+        val dagTemplateName = dagConfig.template
```
tiboun, line 130: Grouping by targetFileName may group different template name because in an autotask, we can override dagRef. We must ensure that the config we use to generate with is the same for all the tasks grouped IMO. If not the case, we should stop the generation and explain why we can't generate.

fupelaqu: We can add this test but I am not sure that it is relevant.

Indeed if the tasks use different dag templates, which may absolutely be the case, it means that in this particular case they will end with dags overwriting the previous ones. 

In other words, it would mean that those different configurations would define exactly the same output filename which should not be the case.

tiboun: Yes, this can happen when user misconfigured dag generation. It allows the user to easily identify this mistake and modify dag generation config accordingly

PR #720, opened 2024-01-09
feat: load gzip compressed files (.gz extension) into bigquery

tiboun:
load gzip compressed files into bigquery.

Tested from local only.

IMO, would be great to remove LocalStorageHandler, is it still usefill to keep it ?


hayssams:
Gr8 PR
I agree that we should remove local storage

PR #733, opened 2024-01-14
update dag generation so as to apply jinja templating on configuration file

fupelaqu:
## Summary
update dag generation so as to apply jinja templating on configuration file when generating each dag file in order to add special dag variables to dag configuration options and comment

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/ai/starlake/schema/handlers/SchemaHandler.scala
+++ b/src/main/scala/ai/starlake/schema/handlers/SchemaHandler.scala
@@ -531,21 +531,7 @@ class SchemaHandler(storage: StorageHandler, cliEnv: Map[String, String] = Map.e
           dagFileContent,
           dagsConfigsPath.toString
         ) match {
-        case Success(dagConfig) =>
-          // we save the raw filename since it willl be instantiaed at runtime (it holds the domain and table vars potentially
-          val rawFilename = dagConfig.filename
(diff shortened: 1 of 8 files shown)


src/main/scala/ai/starlake/schema/handlers/SchemaHandler.scala
```diff
@@ -531,21 +531,7 @@ class SchemaHandler(storage: StorageHandler, cliEnv: Map[String, String] = Map.e
...
-              Utils.parseJinja(dagFileContent, activeEnvVars()),
-              dagsConfigsPath.toString
-            )
-          dagConfigResult match {
-            case Success(dagConfig) =>
-              dagConfig.copy(filename = rawFilename)
-            case Failure(err) =>
-              throw err
-          }
+        case Success(dagConfig) => dagConfig
```
hayssams, line 534: At this step, all variables have been replaced by their values.
How do you detect them when building the dags.
That’s what raw file name was used for previously.

fupelaqu: All variables are replaced in the corresponding dag generation method in DagGenerateCommand

hayssams (approved):
LGTM

PR #736, opened 2024-01-23
Review/starlake airflow documentation

zedach:
## Summary
Review for the starlake-airflow-documentation

**PR Type: Documentation**

**Status: WIP**

### Solution
- fixing some typos
- import link 
- suggest adding Prerequisites check form more clarity to let the user be adequately prepared before installation.
- what happen when StarlakePreLoadStrategy NONE ?
-

### Remaining Todos
for discussion @fupelaqu 
* [ ] - I think we should create a Usage section. that point to other .md in order to make The Readme concise. (especially for  the On premise part)
* [ ] - Showing how to use the module from a starlake perspective (edit yaml file then generate with command, then open generated file, see what is generated before deploying it to airflow, and then show the compiled dag with group task created
* [ ] - unimplemented documentation or feature sould be under To-Do section IMO for transparency and showing current potential. it also acts like project road map


fupelaqu:
> * [ - ]  - I think we should create a Usage section. that point to other .md in order to make The Readme concise. (especially for  the On premise part)

This documentation is aimed at to be packaged directly with the python module. I don't think what you suggest is the most convenient way to do so.

> * [ - ]  - Showing how to use the module from a starlake perspective (edit yaml file then generate with command, then open generated file, see what is generated before deploying it to airflow, and then show the compiled dag with group task created

This part will be integrated within the documentation related to the dag generation. 

As settled as an introduction to the present documentation :
"
starlake-airflow is the [Starlake](https://starlake-ai.github.io/starlake/index.html) Python Distribution for Airflow.

It is recommended to use it in combinaison with [starlake dag generation](https://starlake-ai.github.io/starlake/docs/concepts/orchestration), but can be used directly as is in your DAGs.

" 

> * [x]  - unimplemented documentation or feature should be under To-Do section IMO for transparency and showing current potential. it also acts like project road map

You are absolutely right ! Thanks to point this out

src/main/python/README.md
```diff
@@ -2,7 +2,17 @@
...
-It is recommended to use it in combinaison with **[starlake dag generation](https://starlake-ai.github.io/starlake/docs/concepts/orchestration)**, but can be used directly as is in your **DAGs**.
+It is recommended to use it in combination with  **[starlake dag generation](https://starlake-ai.github.io/starlake/docs/concepts/orchestration)**, but can be used directly as is in your **DAGs**.
+
+## Prerequisites
+
+Before installing `starlake-airflow`, ensure the following minimum versions are installed on your system:
+
+- starlake 1.0.0
+- Python: 3.8 or higher
+- Apache Airflow: 2.4 or higher
```
fupelaqu: Actually to support BashSensor operator with retry_exit_code, Apache Airflow v 2.6 or higher should be required.
For older version, the option retry_on_failure for AirflowStarlakeCloudRunJob will fail

src/main/python/README.md
```diff
@@ -12,7 +22,7 @@ pip install starlake-airflow --upgrade
 
 ## AirflowStarlakeJob
 
-`ai.starlake.job.airflow.AirflowStarlakeJob` is an **abstract factory class** that extends the generic factory interface `ai.starlake.job.IStarlakeJob` and is responsible for **generating** the **Airflow tasks** that will run the `import`, [load](https://starlake-ai.github.io/starlake/docs/concepts/load) and [transform](https://starlake-ai.github.io/starlake/docs/concepts/transform) starlake commands.
+`ai.starlake.job.airflow.AirflowStarlakeJob` is an **abstract factory class** that extends the generic factory interface `ai.starlake.job.IStarlakeJob` and is responsible for **generating** the **Airflow tasks** that will run the [import](https://starlake-ai.github.io/starlake/docs/cli/import) command, the [load](https://starlake-ai.github.io/starlake/docs/concepts/load) and [transform](https://starlake-ai.github.io/starlake/docs/concepts/transform) starlake commands.
```
fupelaqu: The import link should rather point to https://starlake-ai.github.io/starlake/docs/user-guide/load#import-step

src/main/python/README.md
```diff
@@ -77,7 +87,7 @@ def sl_transform(
 
 ### sl_job
 
-Ultimitely, all of these methods will call the `sl_job` method that neeeds to be **implemented** in all **concrete** factory classes.
+Ultimately, all these methods will call the `sl_job` method  method that needs to be **implemented** in all **concrete** factory classes.
```
fupelaqu: The word method is repeated 2 times or a comma is missing :)

src/main/python/README.md
```diff
@@ -132,11 +142,11 @@ def sl_pre_load(
 
 ##### StarlakePreLoadStrategy.NONE
 
-No pre load strategy.
+No pre-load strategy: All loads task group will be directly executed. no pre-load tasks group will be created. include reference to [src/main/python/ai/starlake/job/airflow/airflow_starlake_job.py](https://github.com/zedach/starlake/blob/eb67c427a32222d0b83016f01b18862f4ca49ea4/src/main/python/ai/starlake/job/airflow/airflow_starlake_job.py#L134) TODO include Image.
```
fupelaqu, line 145: It is not totally exact... The pre_tasks method that is called within the pre_load may return a task (at least for Dataproc, there will be a task devoted to the creation of the cluster)

fupelaqu (changes requested):

src/main/python/README.md
zedach: I see. In any case, it was an example I put up to show that we need to talk about it, to put an image so that the user knows what to get in the case of NONE. Typically, it's interesting to know that `pre_load` can return a task in the case of the creation of the spark cluster.

src/main/python/README.md
zedach:
```suggestion
Ultimately, all these methods will call the `sl_job` method that needs to be **implemented** in all **concrete** factory classes.
```

src/main/python/README.md
zedach: I've put the versions for illustration. I've put 2.4 in reference to the introduction of data aware scheduling. 
```suggestion
- Apache Airflow: 2.6 or higher
```

src/main/python/README.md
zedach:
```suggestion
`ai.starlake.job.airflow.AirflowStarlakeJob` is an **abstract factory class** that extends the generic factory interface `ai.starlake.job.IStarlakeJob` and is responsible for **generating** the **Airflow tasks** that will run the [import](https://starlake-ai.github.io/starlake/docs/user-guide/load#import-step) command, the [load](https://starlake-ai.github.io/starlake/docs/concepts/load) and [transform](https://starlake-ai.github.io/starlake/docs/concepts/transform) starlake commands.
```

src/main/python/README.md
zedach: @fupelaqu is it ok for python version?

hayssams:
Closing: 2.7 years old; the airflow documentation has been restructured since.

PR #747, opened 2024-01-30
fix: Align infered schema primitive's type

tiboun:
Align infered schema primitive's type with the one declared `types.sl.yml`.

For instance int was infered as integer primitive type. Making it impossible to ingest the files by default because this primitive type doesn't exist.

### DIFF
--- a/src/main/scala/ai/starlake/schema/handlers/InferSchemaHandler.scala
+++ b/src/main/scala/ai/starlake/schema/handlers/InferSchemaHandler.scala
@@ -48,33 +48,38 @@ object InferSchemaHandler {
 
     schemaWithIndex
       .map { case (row, index) =>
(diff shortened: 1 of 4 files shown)


src/main/scala/ai/starlake/schema/handlers/InferSchemaHandler.scala
```diff
@@ -93,7 +98,7 @@ object InferSchemaHandler {
                 else
                   "timestamp"
               } else
-                row.dataType.typeName
+                PrimitiveType.from(row.dataType).value
```
hayssams, line 101: good catch !

hayssams (approved):
LGTM

hayssams (approved):
LGTM

PR #753, opened 2024-02-16
Fix setup behind proxy

fupelaqu:
## Summary
Fix setup behind proxy

**PR Type: Bug Fix**

**Status: Ready**

**Breaking change? No**


distrib/starlake.sh
```diff
@@ -39,12 +51,36 @@ get_binary_from_url() {
...
 
+add_server_cert_to_java_keystore() {
+  local server=$1
+  local proxy=$2
+  if [ -n "${JAVA_HOME}" ]; then
+    local pem_file="${server}.pem"
+    openssl s_client -proxy "$proxy" -showcerts -servername "$server" -connect "${server}:443" </dev/null | openssl x509 -outform PEM > "$pem_file"
+    local keytool="${JAVA_HOME}/bin/keytool"
+    local alias=$server
+    $keytool -delete -alias "$alias" -keystore "${JAVA_HOME}/lib/security/cacerts" -storepass changeit </dev/null
```
tiboun: changeit est le mot de passe par défaut mais comme son nom l'indique, est-ce que certains clients ne le changerait pas ?

distrib/starlake.sh
```diff
@@ -28,9 +28,21 @@ then
...
+    local server=$1
+    local url=$2
+    local target_file=$3
+    if [[ -n "${https_proxy}" ]] || [[ -n "${http_proxy}" ]]; then
+      if [ -n "${https_proxy}" ]; then
+        local proxy=$https_proxy
+      else
+        local proxy=$http_proxy
+      fi
+      local pem_file="${server}.pem"
```
tiboun, line 40: j'ai l'impression que l'on récupère à chaque fois qu'on lance cette fonction. Si on ajoute le certificat, ne devrait-on pas le supprimer post fonction ?

distrib/starlake.sh
```diff
@@ -39,12 +51,36 @@ get_binary_from_url() {
     fi
 }
 
+add_server_cert_to_java_keystore() {
+  local server=$1
+  local proxy=$2
+  if [ -n "${JAVA_HOME}" ]; then
+    local pem_file="${server}.pem"
```
tiboun, line 69: Idem ici

distrib/starlake.sh
```diff
@@ -53,7 +89,7 @@ launch_setup() {
       exit 1
     fi
   fi
-  $RUNNER -cp "$SCRIPT_DIR/setup.jar" Setup "$SCRIPT_DIR"
+  $RUNNER -cp "$SCRIPT_DIR/setup.jar" Setup "$SCRIPT_DIR" unix
```
tiboun, line 114: Pourquoi avons-nous besoin de ce nouvel argument unix ? Est-ce lié à git bash ?

distrib/starlake.sh
fupelaqu: on n'ajoute pas le certificat à ce niveau, on l'utilise pour l'appel curl

distrib/starlake.sh
fupelaqu: on peut le supprimer mais quelle est la valeur ajoutée ?

distrib/starlake.sh
fupelaqu: c'est lié au fait que si tu lances un setup.sh alors au final tu récupéreras un starlake.sh qui lui même repose sur un versions.sh. Ce paramètre permet de résoudre le cas où tu lances une installation starlake en utilisant les scripts sh depuis un poste windows (via notamment git bash) en générant au final un versions.sh

distrib/starlake.sh
fupelaqu: je n'ai jamais rencontré le cas personnellement mais sinon ils devront modifier le script :)

distrib/starlake.cmd
```diff
@@ -74,7 +92,20 @@ if "%javaVersion%" LSS "11000" (
...
 set setup_url=https://raw.githubusercontent.com/starlake-ai/starlake/master/distrib/setup.jar
-curl -s -o %SCRIPT_DIR%setup.jar %setup_url%
+if defined https_proxy (set "proxy=%https_proxy%") else if defined http_proxy (set "proxy=%http_proxy%") else (echo no proxy)
+if not "%proxy%"=="" (
+    where /q openssl
+    IF ERRORLEVEL 1 (
+        echo openssl is missing. Ensure it is installed and placed in your PATH
+        exit /b
+    )
+    curl.exe -k -s -o %SCRIPT_DIR%setup.jar %setup_url%
```
tiboun: est-ce que cela ne devrait pas être remplacer par un appel vers get_binary_from_url ? est-ce que la commande curl ne supporte pas la notion de proxy sous windows ? le -k est ce que tu voulais éviter de base il me semble.

fupelaqu: curl.exe --cacert %SCRIPT_DIR%raw.githubusercontent.com.pem --proxy %proxy% -s -o %SCRIPT_DIR%setup.jar %setup_url% ne fonctionne pas sous windows (la même commande fonctionne sous Mac / Linux)
j'ai rajouté la possibilité de spécifier la variable d'environnement INSECURE_URL pour forcer le chargement sans vérifier la validité du certificat

hayssams:
@fupelaqu Thanks for submitting this PR.
What you are trying to achieve here is not to make setup work behind a proxy but to bypass the certificate revocation check that may not work in some corporate environments.

Why don't we simply allow the user to explicitly request to bypass this security check by using an env var that would be taken into account by the setup.jar and the setup.sh scripts ?

An implementation would be to add the --ssl-no-revoke argument to the curl command and make the SSL connection trusted in the jar  by defining a custom trust handler or maybe using  System.setProperty("com.sun.net.ssl.checkRevocation", "false");

P.S. We may even display a message like the following to the user:
Certificate revocation is a security feature in which certificates are invalidated before their actual expiration dates. Use this feature at your own risk.

fupelaqu:
@hayssams

My understanding is that the inability to verify certificate revocation is related to the fact that the company's trusted root CA used to sign the server certificate returned by the proxy is not available online. 

On Windows, a successfully tested solution is to use the --ssl-revoke-best-effort flag which ignores the missing/offline certificate CRL.

On Unix both flags (--ssl-no-revoke and  --ssl-revoke-best-effort) have been tested and did not work. 
Nevertheless, a successfully tested solution is to download the certificate using openssl and use it in conjunction with the curl -cacert flag.

Regarding Java, I don't think it's a good idea to define a custom certificate trust handler as we discussed recently and the com.sun.net.ssl.checkRevocation option is not working. 

The solution that was successfully tested is to download the certificates from the server using openssl again and add them to the local Java keystore. 

Using openssl to download certificates doesn't mean that we can't validate them. On the opposite, by using verify_return_error flag, we can guarantee that the certificate has been checked.
However, in this case, the user will have to specify the path to the offline CA which can be done whether by specifying the path to a CA pem file or directory. 

However, I agree that we should add a variable to let the user decide whether or not they want to bypass these security controls.

I will update the PR accordingly.

hayssams:
Do you think that it would be possible to remove the burden for the user to specify a CA path.
On my side I tried using curl on git bash on Windows and it worked. 

Regarding the TrustManager, if we activate the custom Trust Manager only when the user request an insecure connection then your previous solution makes sense and would removed the burden from the user.
What do you think ?

hayssams:
Also to avoid having Setup$1.class generated can't we make the Setup class itself directly inherit from the X598 TrustManager class ?

fupelaqu:
The CA path is not required on Windows to download binary files. It is required only if you are behind a proxy to download the certificates that will be added to the local Java keystore (if SL_INSECURE has not been set).

If SL_INSECURE has been set, the certificates will be downloaded directly without having to specify the CA path.

distrib/setup.sh
```diff
@@ -31,8 +31,31 @@ get_installation_directory() {
...
+    if [[ -n "${https_proxy}" ]] || [[ -n "${http_proxy}" ]]; then
+      if [ -n "${https_proxy}" ]; then
+        local proxy=$https_proxy
+      else
+        local proxy=$http_proxy
+      fi
+      local pem_file="${server}.pem"
+      if [ -n "${SL_INSECURE}" ]; then
+        openssl s_client -proxy "$proxy" -showcerts -servername "$server" -connect "${server}:443" </dev/null 2>/dev/null | openssl x509 -outform PEM > "$pem_file" 2>/dev/null
+      else
```
hayssams, line 45: Why do we need to use OpenSSL and CA Path if the user did not request an insecure path ?

distrib/starlake.cmd
```diff
@@ -57,6 +57,39 @@ if "%~1"=="install" (
...
+    if defined SL_INSECURE (
+        openssl s_client -proxy %proxy% -showcerts -servername %~1 -connect %~1:443 > %~1.tmp 2>nul
+    ) else if defined CA_PATH (
+        echo Verifying %~1 certificate using CApath %CA_PATH%...
+        openssl s_client -CApath %CA_PATH% -proxy %proxy% -verify_return_error -showcerts -servername %~1 -connect %~1:443 > %~1.tmp 2>nul
+    ) else if DEFINED CA_FILE (
+        echo Verifying %~1 certificate using CAfile %CA_FILE%...
+        openssl s_client -CAfile %CA_FILE% -proxy %proxy% -verify_return_error -showcerts -servername %~1 -connect %~1:443 > %~1.tmp 2>nul
+    ) else (
+        echo Retrieving %~1 certificate...
```
hayssams, line 71: Why ...

distrib/starlake.cmd
```diff
@@ -74,7 +107,21 @@ if "%javaVersion%" LSS "11000" (
 )
 
 set setup_url=https://raw.githubusercontent.com/starlake-ai/starlake/master/distrib/setup.jar
-curl -s -o %SCRIPT_DIR%setup.jar %setup_url%
+if defined https_proxy (set "proxy=%https_proxy%") else if defined http_proxy (set "proxy=%http_proxy%") else (echo no proxy)
+if not "%proxy%"=="" (
```
hayssams, line 111: The script does not work if openssl is not installed. This makes openssl a requirement

fupelaqu:
The current solution that has been implemented is the best solution for me. 

It works as is without having to define a custom TrustManager, and specifying SL_INSECURE is sufficient for the user to install starlake without having to specify other varaibles.

distrib/starlake.cmd
fupelaqu: it is required only if a proxy has been found...

The script simply doesn't work otherwise

distrib/setup.sh
fupelaqu: It is inexact. If SL_INSECURE has been specified, no CA path is required

distrib/starlake.cmd
fupelaqu: because in this case, SL_INSECURE has not been defined

distrib/setup.sh
hayssams: I see that between line 43 and line 51 openssl appears in all paths. Which means that openssl is always invoked even if not necessary

distrib/starlake.cmd
hayssams: Why do we need to call openssl when SL_INSECURE is not defined ?

distrib/starlake.cmd
hayssams: I had no issue when behind a proxy from my office with the current script

hayssams:
Yes but it requires to install openssl just for downloading starlake ...

distrib/setup.sh
fupelaqu: Well it didn't work for me at all without

distrib/starlake.cmd
fupelaqu: Well maybe this PR should be closed :) except for this 2 fixes : 

- when using a proxy which does not require any authentication (Setup.java line 55 to 61)
- when installed using starlake.sh from Windows (using git bash)

fupelaqu:
At least it doesn't require a custom TrustManager :D

hayssams:
The TrustManager would be used only if the user request it using the enc var sl_insecure


I find it intrusive to update the user's key store only to download a jar artefact

Actually I agree that this PR should not go in the main branch but probably appear in a contrib section in a specific folder with a readme explaining the problem it addresses.

PR #766, opened 2024-02-21
update starlake install

fupelaqu:
## Summary
This update to fix 2 issues with starlake install:

- unable to install starlake when using a proxy which does not require authentication
- unable to launch starlake after an install with setup.sh on Windows (through git bash for instance)

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


src/main/java/Setup.java
```diff
@@ -51,14 +51,12 @@ private static void parseProxy(String proxy) {
             } else {
                 host = hostAndPort;
             }
+        } else if (proxy.contains(":")) {
```
hayssams, line 54: this condition is always true

hayssams: how will it work with proxy=http://myproxy:8080

fupelaqu: this condition is not always true (proxy without port and scheme)

you are right it will not work in this case but it doesn't work when the proxy is defined for instance as 127.0.0.1:3128

we have to deal with both cases

hayssams: you cannot set a proxy without the schemeBecause we may have socks:// .....
That's why the protocol is always required or else we should have another variable that define tell about the type of proxy http/https/socks ...

fupelaqu: I don't see in the code that it is required, should we throw an exception in this case ?

fupelaqu: I may be wrong but if the scheme is socks, should not we define socksProxyHost and socksProxyPort instead of respectively http(s).proxyHost and http(s).proxyPort ?

hayssams: You’re right about the socks stuff.

The setup code never check if the url starts with http or https. It is a good improvement to raise an error if the user has not set correctly the proxy url.

hayssams (approved):
LGTM

PR #771, opened 2024-02-21
feat: add hash or timestamp to snapshot version

tiboun:
add additional information for SNAPSHOT version during runtime in order to know more precisely which version we are running while deploying a SNAPSHOT.


hayssams (approved):
LGTM

PR #776, opened 2024-02-22
add SL_INSECURE for Setup behind proxy

fupelaqu:
## Summary
Sometimes, when using a corporate proxy, the certificate revocation check is not possible because the company's trusted root CA used to sign the server certificate returned by the proxy is not available online.

To bypass this limitation, this PR introduces SL_INSECURE environment variable which allows insecure server connections.

Additionally, this PR adds support for SOCKS type proxy during the installation.

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


src/main/java/Setup.java
```diff
@@ -40,59 +55,54 @@ private static void parseProxy(String proxy) {
...
-                final String[] hostAndPort = schemeWithHostAndPort[1].split(":");
+        final Pattern pattern = Pattern.compile("(https?|socks5?):\\/\\/([^:].+)", Pattern.CASE_INSENSITIVE);
+        final Matcher m = pattern.matcher(proxy);
+        if (m.matches()) {
+            protocol = m.group(1).toLowerCase();
+            final String hostAndPortWithMaybeCredentials = m.group(2);
+            if (hostAndPortWithMaybeCredentials.contains("@")) {
+                final String[] hostAndPortWithCredentials = hostAndPortWithMaybeCredentials.split("@");
+                final String[] credentials = hostAndPortWithCredentials[0].split(":");
+                username = credentials[0];
```
tiboun, line 67: I suggest to add an assert requiring a length of two and inform to the user that the credential is expecting a separator with `:`

Otherwise, we have to change the regex and makes sure that it contains a `:` before @.

distrib/setup.sh
```diff
@@ -30,9 +30,18 @@ get_installation_directory() {
...
 }
 
+if [[ -n "${https_proxy}" ]] || [[ -n "${http_proxy}" ]]; then
+  PROXY=${https_proxy:-$http_proxy}
+fi
+
 get_from_url() {
     local url=$1
-    local response=$(curl -s -w "%{http_code}" "$url")
+    if [ -n "$PROXY" ] && [ -n "$SL_INSECURE" ]; then
```
tiboun, line 39: does proxy always requires to be used with insecure ?

fupelaqu: it is the aim of this PR, if SL_INSECURE flag has been specified and a proxy has been defined, then we add --insecure flag to curl. If either proxy or SL_INSECURE flag has not been defined, then we use curl without insecure flag.

src/main/java/Setup.java
fupelaqu: It has been taken into account by enforcing credentials requirements
```java
assert(credentials.length == 2): "Invalid credentials format, expecting 'username:password'";
```

src/main/java/Setup.java
```diff
@@ -284,6 +295,61 @@ private static boolean anyDependencyEnabled() {
...
+    private static final TrustManager alwaysTrustManager = instance;
+
+    private static final ProxySelector proxySelector = instance;
+
+    private static void setHttpClient() throws NoSuchAlgorithmException, KeyManagementException {
+        setProxy();
+        HttpClient.Builder clientBuilder = HttpClient.newBuilder();
+        clientBuilder.proxy(proxySelector);
+        if (username != null && password != null) {
+            Authenticator authenticator = new Authenticator() {
```
hayssams, line 332: doesn't this call create a new . class file ? If so can we make setup inherit Authenticator ?

src/main/java/Setup.java
```diff
@@ -441,7 +506,7 @@ private static void downloadAndDisplayProgress(String urlStr, String file) throw
                 StringBuilder sb = new StringBuilder("Progress: " + (total / 1024 / 1024) + "/" + (lengthOfFile / 1024 / 1024) + " MB");
                 if (lengthOfFile > 0) {
                     sb.append(" (");
-                    sb.append((int) (total * 100 / lengthOfFile));
+                    sb.append((total * 100 / lengthOfFile));
```
hayssams: remove extra parentheses

hayssams (approved):
LGTM

src/main/java/Setup.java
fupelaqu: Yes but unfortunately no choice because Authenticator is an abstract Class and Setup extends already an abstract class (ProxySelector)

src/main/java/Setup.java
fupelaqu: Done

PR #782, opened 2024-02-24
Allow the user to restrict the rows to extract

hayssams:
## Summary
A new attribute ("where") has been added and is applied on extraction.

**PR Type:  Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/ai/starlake/extract/JDBCSchemas.scala
+++ b/src/main/scala/ai/starlake/extract/JDBCSchemas.scala
@@ -97,7 +97,7 @@ case class JDBCSchema(
   stringPartitionFunc: Option[String] = None,
   fullExport: Option[Boolean] = None,
   sanitizeName: Option[Boolean] = None,
-  where: Option[String] = None
+  filter: Option[String] = None
 ) {
   def this() = this(None) // Should never be called. Here for Jackson deserialization only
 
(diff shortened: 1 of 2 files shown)


src/main/scala/ai/starlake/extract/JDBCSchemas.scala
```diff
@@ -96,7 +96,8 @@ case class JDBCSchema(
   fetchSize: Option[Int] = None,
   stringPartitionFunc: Option[String] = None,
   fullExport: Option[Boolean] = None,
-  sanitizeName: Option[Boolean] = None
+  sanitizeName: Option[Boolean] = None,
+  where: Option[String] = None
```
tiboun: In load, I think we named this attribute `filter` and not `where` and they have the same feature.

src/main/scala/ai/starlake/extract/JdbcDbUtils.scala
```diff
@@ -1278,11 +1278,12 @@ object JdbcDbUtils extends LazyLogging {
     auditColumns: Columns
   )(implicit settings: Settings): Try[Unit] = {
     val dataColumnsProjection = tableExtractDataConfig.columnsProjectionQuery(extractConfig.data)
+    val where = extractConfig.jdbcSchema.where.map(w => s"where $w").getOrElse("")
```
tiboun: We should rename it to extraCondition or rename above one to where in order to have the same name for both function

src/main/scala/ai/starlake/extract/JDBCSchemas.scala
hayssams: You're right. 
Change made

src/main/scala/ai/starlake/extract/JdbcDbUtils.scala
hayssams: Done

tiboun:
The internal boundaries does apply filter in order to compute them. The output is the same but may implies more executed query

PR #783, opened 2024-02-24
Revert "Allow the user to restrict the rows to extract"

hayssams:
Reverts starlake-ai/starlake#782


hayssams:
@tiboun could you explain a little bit more => 
The internal boundaries does apply filter in order to compute them. The output is the same but may implies more executed query

hayssams:
Understood. Just made the correction

PR #785, opened 2024-02-26
update xls2yml adding columns

fupelaqu:
## Summary

Through xls2yml :

- add specified tags to the domain (useful not to duplicate the same tags for all domain's schemas).
- add optional tags to the auto job
- add optional clustering to the auto job

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


src/main/scala/ai/starlake/schema/generator/XlsModel.scala
```diff
@@ -12,6 +12,7 @@ trait XlsModel {
     "_path"        -> "Directory",
     "_ack"         -> "Ack",
     "_description" -> "Description",
+    "_tags"        -> "Tags",
```
hayssams: doesn't it break existing xls if we do not put the new attribute at the end  ?

hayssams (changes requested):

src/main/scala/ai/starlake/schema/generator/XlsModel.scala
fupelaqu: yep, I just fixed it

hayssams (approved):
LGTM

PR #788, opened 2024-02-28
feat: validate structure with json-schema

tiboun:
Validation config structure against json schema


src/main/scala/ai/starlake/schema/model/Domain.scala
```diff
@@ -82,7 +82,7 @@ case class LoadDesc(load: Domain)
   *   specified in the settings. This attribute is optional.
   */
 @nowarn case class Domain(
-  name: String,
+  name: String, // TODO: should we remove it from json schema and apply a not required on it?
```
hayssams: The name is already not required 
But we need it in the json schema since the user is allowed to have a name different from the folder name

src/main/scala/ai/starlake/schema/model/Metadata.scala
```diff
@@ -192,17 +193,15 @@ case class Metadata(
...
   def getNullValue(): String = nullValue.getOrElse(if (isEmptyIsNull()) "" else null)
   // scalastyle:on null
 
   @JsonIgnore
-  def getPartitionAttributes()(implicit settings: Settings): List[String] = {
+  def getPartitionAttributes()(implicit settings: Settings): Seq[String] = {
     this.getSink().toAllSinks().partition.getOrElse(Nil)
   }
 
-  @JsonIgnore
```
hayssams, line 205: We need to kee this or else Jackson on write would generate this field

hayssams (approved):
Gr8 PR
We should just keep the jsonignore annotation

src/main/scala/ai/starlake/schema/model/Domain.scala
tiboun: It's right that the user can have a name different from the file one but when we restrict on the domain, we only check if the directory has the same name. Therefore, I think what we got the different scenarios:
- user load all domain:
    - if the domain doesn't have a rename, then the given name is used as target dataset|schema name. This is equivalent to have the name equal to the folder name and have a rename attribute if the name was different from the folder name
    - otherwise if a rename is there, I think we don't use it to filter or compute anything.
- user load one specific domain: 
    - if the given name is equal to the folder name, it works
    - if the given name differs from the folder name, we have two cases:
       - if the user includes the domain based on the name in the config file, we state that there is no domain to load. We don't fallback to read all config files as well because what we want when we filter domains is to have avoid the parsing of all files. The user has to change the name equal to the folder and apply a rename to it in order to use the filtering. It has to use the folder name to ingest.
       -  if the user includes the domain based on the folder name, we load the file and it works

IMO, we can keep the name, I will remove the TODO but what we should do is to add a validation rule on it and check that the name if defined is equal to the folder and if rename is not defined, ask the user to define it with the current value.

What do you think ?

src/main/scala/ai/starlake/schema/model/Metadata.scala
tiboun: As discussed, removal of JsonIgnore is kept because it is a value that can be filled by the end user. If we don't plan to change the client value, I would suggest to rename all fields acting like this (and I think there is some more) as getEffective* with a @JsonIgnore on it.

Therefore, we don't change the user input on serialization but we get the value to use in our process. What do you think ?

PR #801, opened 2024-03-08
update Xls Domain Reader to define quote and nullValue

fupelaqu:
## Summary
update Xls Domain Reader to define quote and nullValue within schema metadata
allow empty value for quote
update default value for quote as empty string

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

PR #803, opened 2024-03-08
Fix/default quote

fupelaqu:
## Summary
revert to previous default quote
update xls so as to allow blank quote using "blank" within shema metadata

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

PR #883, opened 2024-05-17
take into account optional cron within datasets

fupelaqu:
## Summary
The goal of this PR is to support data lineage freshness in dags generated by Starlake for Airflow.

Data lineage is implemented within Airflow by using data-aware scheduling.

Each time a load or transformation is executed within a dag, Starlake will produce the corresponding Dataset using outlets.

Similarly, when a transformation requires that other load(s) or transformation(s) be performed beforehand, Starlake will add them as a list of Datasets to the corresponding consumer dag so that the dag will be automatically triggered each time all the datasets would have been produced .

The problems start when a dag has dependencies on Datasets that are produced by dag(s) whose frequency(ies) may differ. 

For example a dag may depend on one Dataset A that is produced once a month and another Dataset B that is  produced daily... 
At the end the dag will be triggered not every day as expected but not before the dag that produces Dataset A will be triggered, which means once a month.

The freshness of the data will ultimately be defined by the dataset updated the least frequently.

The solution consists of ensuring that all datasets produced by a dag for which a cron has been defined are republished identically by the consumer dag.

This solution implies : 

- adding the optional schedule (cron) for each dependency computed for dags
- adding to the uri of each dataset produced by a dag for which a cron has been defined, in the form of a query parameter (sl_schedule), the logical date time of the dag
- adding to the uri of each dataset consumed by a dag for which the corresponding dependency has defined a cron, in the form of a query parameter (sl_schedule), the last logical date time the dataset should have been produced
- republishing those scheduled datasets

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_cloud_run_job.py
+++ b/src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_cloud_run_job.py
@@ -26,6 +26,7 @@ class StarlakeAirflowCloudRunJob(StarlakeAirflowJob):
             cloud_run_job_region: str=None,
             options: dict=None,
             cloud_run_async:bool=None,
+            cloud_run_async_poke_interval: float=None,
             retry_on_failure: bool=None,
             retry_delay_in_seconds: float=None,
             separator:str = ' ',
@@ -35,6 +36,7 @@ class StarlakeAirflowCloudRunJob(StarlakeAirflowJob):
         self.cloud_run_job_name = __class__.get_context_var(var_name='cloud_run_job_name', options=self.options) if not cloud_run_job_name else cloud_run_job_name
         self.cloud_run_job_region = __class__.get_context_var('cloud_run_job_region', "europe-west1", self.options) if not cloud_run_job_region else cloud_run_job_region
         self.cloud_run_async = __class__.get_context_var(var_name='cloud_run_async', default_value="True", options=self.options).lower() == "true" if cloud_run_async is None else cloud_run_async
+        self.cloud_run_async_poke_interval = float(__class__.get_context_var('cloud_run_async_poke_interval', "30", self.options)) if not cloud_run_async_poke_interval else cloud_run_async_poke_interval
         self.separator = separator if separator != ',' else ' '
         self.update_env_vars = self.separator.join([(f"--update-env-vars \"^{self.separator}^" if i == 0 else "") + f"{key}={value}" for i, (key, value) in enumerate(self.sl_env_vars.items())]) + "\""
         self.retry_on_failure = __class__.get_context_var("retry_on_failure", "False", self.options).lower() == 'true' if retry_on_failure is None else retry_on_failure
@@ -50,7 +52,7 @@ class StarlakeAirflowCloudRunJob(StarlakeAirflowJob):
                     f"gcloud beta run jobs execute {self.cloud_run_job_name} "
                     f"--args \"{command}\" "
                     f"{self.update_env_vars} "
-                    f"--async --region {self.cloud_run_job_region} --project {self.project_id} --format='get(metadata.name)'" #--task-timeout 300 
+                    f"--async --region {self.cloud_run_job_region} --project {self.project_id} --format='get(metadata.name)'" #--task-timeout 300
                 ),
                 do_xcom_push=True,
                 **kwargs
@@ -63,6 +65,7 @@ class StarlakeAirflowCloudRunJob(StarlakeAirflowJob):
                 cloud_run_job_region=self.cloud_run_job_region,
                 source_task_id=job_task.task_id,
                 retry_on_failure=self.retry_on_failure,
+                poke_interval=self.cloud_run_async_poke_interval,
                 **kwargs
             )
             if self.retry_on_failure:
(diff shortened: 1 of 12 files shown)


mhdkassir (approved):
Lgtm

src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_cloud_run_job.py
```diff
@@ -35,6 +36,7 @@ def __init__(
         self.cloud_run_job_name = __class__.get_context_var(var_name='cloud_run_job_name', options=self.options) if not cloud_run_job_name else cloud_run_job_name
         self.cloud_run_job_region = __class__.get_context_var('cloud_run_job_region', "europe-west1", self.options) if not cloud_run_job_region else cloud_run_job_region
         self.cloud_run_async = __class__.get_context_var(var_name='cloud_run_async', default_value="True", options=self.options).lower() == "true" if cloud_run_async is None else cloud_run_async
+        self.cloud_run_async_poke_interval = float(__class__.get_context_var('cloud_run_async_poke_interval', "30", self.options)) if not cloud_run_async_poke_interval else cloud_run_async_poke_interval
```
zedach, line 39: this changes seems to be related to cloud run jobs feature, I'm I wrong ?

src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_cloud_run_job.py
```diff
@@ -50,7 +52,7 @@ def __job_with_completion_sensors__(self, task_id: str, command: str, spark_conf
                     f"gcloud beta run jobs execute {self.cloud_run_job_name} "
                     f"--args \"{command}\" "
                     f"{self.update_env_vars} "
-                    f"--async --region {self.cloud_run_job_region} --project {self.project_id} --format='get(metadata.name)'" #--task-timeout 300 
+                    f"--async --region {self.cloud_run_job_region} --project {self.project_id} --format='get(metadata.name)'" #--task-timeout 300
```
zedach, line 55: also Here. is this related to optional cron within dataset ?

src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_dataproc_job.py
```diff
@@ -64,29 +64,31 @@ def __init__(self, cluster_config: StarlakeAirflowDataprocClusterConfig, options
...
+            task_id: str=None,
+            cluster_name: str=None,
+            **kwargs) -> BaseOperator:
         """
         Create the Cloud Dataproc cluster.
         This operator will be flagged a success if the cluster by this name already exists.
         """
         cluster_id = self.cluster_config.cluster_id if not cluster_id else cluster_id
-        cluster_name = f"{self.cluster_config.dataproc_name}-{cluster_id.replace('_', '-')}-{TODAY}" if not cluster_name else cluster_name
+        cluster_name = f"{self.cluster_config.dataproc_name}-{cluster_id.replace('_', '-')}-{TODAY}"[0:51] if not cluster_name else cluster_name[0:51]
```
zedach, line 89: Same here in this file: change seems not to be related to the feature but with dataproc cluster name limitation. maybe should be in a fix PR, what do you think ?

src/main/resources/templates/dags/__common_airflow.py.j2
```diff
@@ -22,3 +22,23 @@ def default_spark_config(*args, **kwargs) -> StarlakeSparkConfig:
...
+    if parameters.__len__() > 0 :
+        return '?' + '&'.join(list(f'{parse.quote(k)}={parse.quote(v)}' for (k,v) in parameters.items()))
+    else :
+        return ''
+
+from croniter import croniter
+from datetime import datetime as dt
+
+def cron_start_time() -> dt:
+    return dt.fromtimestamp(dt.now().timestamp())
```
zedach, line 38: Why not using `dt.now()`  directly to obtain the datetime object ?

src/main/resources/templates/dags/__common_airflow.py.j2
```diff
@@ -22,3 +22,23 @@ def default_spark_config(*args, **kwargs) -> StarlakeSparkConfig:
         **kwargs
     )
 spark_config = getattr(sys.modules[__name__], "get_spark_config", default_spark_config)
+
+from urllib import parse
```
zedach, line 26: maybe import quote directly simplify the code 
`from urllib.parse import quote`

zedach (changes requested):
LGTM. Some changes are due to formatting and other corrections that don't seem to be related to the new functionality. I propose to include them in another PR if that's the case

zedach (approved):

src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_cloud_run_job.py
fupelaqu: it is right but it has been added in this PR while testing the latter

src/main/python/starlake-airflow/ai/starlake/airflow/gcp/starlake_airflow_dataproc_job.py
fupelaqu: You are totally right this update is related with dataproc cluster nae limitation and has been added in this PR while testing the latter. I don't think we should create a specific PR for this

PR #891, opened 2024-05-25
Fix timestamp with timezone

hayssams:


tiboun:
LGTM, so it seems that java.sql.timestamp is affected by the current timezone and when we get the value from a LocalDate, it doesn't. Am I right ? If so good to know !

I've tried this as well `PrimitiveType.timestamp.fromString("2011-12-03T10:15:30+01:00[Europe/Paris]", "ISO_ZONED_DATE_TIME", "UTC")` and it works like a charm, timestamp is given in UTC as expected.

Thanks @hayssams

PR #899, opened 2024-05-27
fix: avoid null pointer during schema inference of columns and handle complex json inference

tiboun:
Infer schema for any json


hayssams (approved):
LGTM
Gr8 PR

PR #909, opened 2024-05-30
chore: tables attached to domain description in load stage is not authorized starting from 1.2.0 but a config migration is automatically done  ## Summary

tiboun:


src/main/scala/ai/starlake/schema/model/Domain.scala
```diff
@@ -231,6 +234,7 @@ case class LoadDesc(version: Int, load: Domain)
 
   def asDot(includeAllAttrs: Boolean, fkTables: Set[String]): String = {
     tables
+      .sortBy(_.name)
```
hayssams, line 237: sortBy(_.getFinaleName) probably

hayssams (approved):
LGTM
Gr8 PR

src/main/scala/ai/starlake/schema/model/Domain.scala
tiboun: This has no importance regarding rendering I think. It was introduced here in order to have an determinist output for unit test assertion.

PR #1014, opened 2024-09-12
fix: fix usage of technical columns when load table is set to overwrite

tiboun:
fix usage of technical columns when load table is set to overwrite


hayssams (approved):
Good catch
LGTM

PR #1027, opened 2024-09-26
fix: json tree validation

tiboun:
This PR fixes tree validation:
  - invalid name comparison
  - fix index increment based on the resulting name comparison
  - struct context was invalid in the error built


hayssams (approved):
LGTM

PR #1039, opened 2024-10-03
add support for loading sql scripts for duckdb within each unit test

fupelaqu:
## Summary
add support for loading sql scripts for duckdb within each unit test

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**

### DIFF
--- a/src/main/scala/ai/starlake/tests/StarlakeTests.scala
+++ b/src/main/scala/ai/starlake/tests/StarlakeTests.scala
@@ -365,7 +365,8 @@ object StarlakeTestData {
         )
       val storage = settings.storageHandler()
       val testEnvVars =
-        EnvDesc.loadEnv(testEnvPath)(storage)
+        EnvDesc
+          .loadEnv(testEnvPath)(storage)
           .map(_.env)
           .getOrElse(Map.empty)
 
@@ -433,7 +434,8 @@ object StarlakeTestData {
         )
       val storage = settings.storageHandler()
       val testEnvVars =
-        EnvDesc.loadEnv(testEnvPath)(storage)
+        EnvDesc
+          .loadEnv(testEnvPath)(storage)
           .map(_.env)
           .getOrElse(Map.empty)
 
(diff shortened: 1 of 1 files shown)


src/main/scala/ai/starlake/tests/StarlakeTests.scala
```diff
@@ -788,6 +804,23 @@ object StarlakeTestData {
...
+    val preTestStatementsFiles = testFolder
+      .listFiles(f => f.isFile && !f.getName.startsWith("_") && f.getName.endsWith(".sql"))
+      .toList
+    val preTestStatements: List[String] =
+      preTestStatementsFiles.flatMap { preTestStatementsFile =>
+        if (preTestStatementsFile.exists()) {
+          val source = Source.fromFile(preTestStatementsFile)
+          val sql = source.mkString
+          source.close()
+          sql.split(";")
```
hayssams: I am not sure that you need to split on ';' and the way it is done is not very safe.
I think that you can just submit all of them

src/main/scala/ai/starlake/tests/StarlakeTests.scala
```diff
@@ -64,6 +65,19 @@ case class StarlakeTest(
       d.load(conn)
     }
     expectations.foreach(_.load(conn))
+    preTestStatements.foreach { sql =>
```
hayssams: No need to split the query.
The code below works well:
```
object X {
  def main(args: Array[String]) = {
    val url = "jdbc:duckdb:/Users/hayssams/tmp/x.db"
    val driver = "org.duckdb.DuckDBDriver"
    val properties = new Properties()
    // properties.put("driver", "org.duckdb.DuckDBDriver")
    val connection = DriverManager.getConnection(url, properties)
    val statement = connection.createStatement()
    statement.execute(
      """
         |CREATE TABLE people (id INTEGER, name VARCHAR);
         |INSERT INTO people VALUES (1, 'Mark'), (2, 'Hannes');""".stripMargin
    )
    statement.close()
    connection.close()

  }
}
```

hayssams (changes requested):

hayssams (approved):
LGTM

PR #1040, opened 2024-10-03
to fix unit test reports generation when outputDir option is specified

fupelaqu:
## Summary
to fix unit test reports generation when outputDir option is specified

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

PR #1063, opened 2024-10-15
add partitioning + clustering for BQ when table does not exist

fupelaqu:
## Summary
Add partitioning and clustering for BQ when table does not exist

**PR Type: Bug Fix**

**Status: WIP**

**Breaking change? No**


hayssams (approved):
LGTM

PR #1064, opened 2024-10-15
fix template resolution && merge statement

tiboun:
When starlake is trying to load from absolute path, if the path doesn't exists, an exception is thrown but not catched. This fix makes it able to fallback to other resolution.

During merge statement, update clause set the different columns but doesn't try to unescape column before escaping it again.


hayssams (approved):
LGTM

PR #1071, opened 2024-10-17
fix starlake commands launch within starlake airflow using sl_en_var option

fupelaqu:
## Summary
fix starlake commands launch within starlake airflow using sl_en_var option

**PR Type: Bug Fix**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

PR #1085, opened 2024-10-22
define each grantee with potential multiple values

fupelaqu:
(usefull when a grantee is defined using an external variable, eg terraform variable)

## Summary
Using excel configuration files, when defining policies whose user groups rely on external variables, after applying xls2yml we will end up with something like : 

```
  acl:
  - role: "roles/bigquery.dataViewer"
    grants:
    - "${my_members}"
    - "user:another_member"
#...
```
This PR is to ensure that, in this case, all the members will be accordingly taken into account

**PR Type: Feature**

**Status: Ready to review**

**Breaking change? No**


hayssams (approved):
LGTM

Issue #1087, opened 2024-10-23
[BUG] - extract from duckdb (tutorial) not working with error  java.sql.SQLException: Connection Error: Can't open a connection to same database 


zedach:
## Description
extract from duckdb (tutorial) not working with error  Exception in thread "main" java.sql.SQLException: Connection Error: Can't open a connection to same database file with a different configuration than existing connections. connections are not closing correctly. 

## Expected behavior
extraction work properly with files extracted for duckdb

## Current Behavior
A clear description of what the bug is.
What is affected? Who is it affecting? When does it occur? Where does it happen?

## Steps to reproduce
1. download [duckdb.db sample ](https://github.com/starlake-ai/starlake/blob/master/samples/starbake/datasets/duckdb.db)
2. configure an extract file 
3. launch `starlake extract-data --config metadata/extract/my_extract_config.sl.yml --outputDir datasets/incoming/starbake `
5. See error
java.sql.SQLException: Connection Error: Can't open a connection to same database file with a different configuration than existing connections
        at org.duckdb.DuckDBNative.duckdb_jdbc_startup(Native Method)
        at org.duckdb.DuckDBConnection.newConnection(DuckDBConnection.java:51)
        at org.duckdb.DuckDBDriver.connect(DuckDBDriver.java:41)
        at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:681)
        at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:190)
        at ai.starlake.extract.JdbcDbUtils$StarlakeConnectionPool$.getConnection(JdbcDbUtils.scala:66)
        at ai.starlake.extract.JdbcDbUtils$.$anonfun$withJDBCConnection$1(JdbcDbUtils.scala:105)
        at scala.util.Try$.apply(Try.scala:217)
        at ai.starlake.extract.JdbcDbUtils$.withJDBCConnection(JdbcDbUtils.scala:105)
        at ai.starlake.extract.ExtractDataJob.$anonfun$extractTableData$4(ExtractDataJob.scala:522)
        at scala.util.Using$.resource(Using.scala:296)
        at ai.starlake.extract.ExtractDataJob.$anonfun$extractTableData$3(ExtractDataJob.scala:469)
        at ai.starlake.extract.JdbcDbUtils$.$anonfun$withJDBCConnection$2(JdbcDbUtils.scala:111)

## Context
Add any other context about the problem here: environment, custom data or schema, version...

# Life cycle

[...]

PR #1088, opened 2024-10-23
 [BUG] - extract from duckdb (tutorial) not working with error java.sql.SQLException: Connection Error: Can't open a connection to same database #1087 

zedach:
## Summary
extract from duckdb (tutorial) not working with error  Exception in thread "main" java.sql.SQLException: Connection Error: Can't open a connection to same database file with a different configuration than existing connections. connections are not closing correctly. 

**Related Issue: #IssueId**

**PR Type: Bug Fix |**

**Breaking change? Yes/No**

### Solution

we can enforce readonly and permit RW connection only on audit insertions with proper close connection.

### DIFF
--- a/src/main/scala/ai/starlake/extract/ExtractDataJob.scala
+++ b/src/main/scala/ai/starlake/extract/ExtractDataJob.scala
@@ -777,7 +777,7 @@ class ExtractDataJob(schemaHandler: SchemaHandler) extends Extract with LazyLogg
           val boundaries = if (extractConfig.data.options == extractConfig.audit.options) {
             getBoundariesWith(connection)
           } else {
-            withJDBCConnection(extractConfig.audit.options) { auditConnection =>
+            withJDBCConnection(readOnlyConnection(extractConfig.audit.options)) { auditConnection =>
               getBoundariesWith(auditConnection)
             }
           }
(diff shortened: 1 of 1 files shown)


src/main/scala/ai/starlake/extract/ExtractDataJob.scala
```diff
@@ -157,7 +157,7 @@ class ExtractDataJob(schemaHandler: SchemaHandler) extends Extract with LazyLogg
       val selectedTablesAndColumns: Map[TableName, ExtractTableAttributes] =
         JdbcDbUtils.extractJDBCTables(
           filteredJdbcSchema.copy(exclude = extractConfig.excludeTables.toList),
-          extractConfig.data,
+          readOnlyConnection(extractConfig.data),
```
hayssams, line 160: useless since already done inside extractJDBCTables.

src/main/scala/ai/starlake/extract/ExtractDataJob.scala
```diff
@@ -777,7 +777,7 @@ class ExtractDataJob(schemaHandler: SchemaHandler) extends Extract with LazyLogg
           val boundaries = if (extractConfig.data.options == extractConfig.audit.options) {
             getBoundariesWith(connection)
           } else {
-            withJDBCConnection(extractConfig.audit.options) { auditConnection =>
+            withJDBCConnection(readOnlyConnection(extractConfig.audit.options)) { auditConnection =>
```
hayssams, line 780: Added this one too

hayssams (approved):
LGTM

PR #1209, opened 2025-02-10
fix: remove extra newline in setup.sh causing incorrect version list

FlavRei:


hayssams (approved):
gr8 catch
Thanks for this fix
LGTM

PR #1226, opened 2025-02-24
add confluent setup that can be enabled via ENABLE_KAFKA

tiboun:
add confluent setup that can be enabled via ENABLE_KAFKA


hayssams (approved):
LGTM

PR #1289, opened 2025-03-27
feat: add encoding option to schema inference

tiboun:
Currently, infer-schema suppose input file to be encoded as utf-8, making it impossible to infer some files when encoding is like cp1252.

Add encoding option to infer-schema command and allow users to change encoding.


hayssams (approved):
LGTM

PR #1290, opened 2025-03-27
fix: temporary table are now set with expiration time

tiboun:
There was a conflict regarding cliConfig.days usage and temporary tables were set with expiration time when partitioning was set. Therefore it now has a dedicated field.


hayssams (approved):
LGTM

PR #1297, opened 2025-03-31
feat: add rawDomains in order to have access to the whole config during templating

tiboun:
Add ability to use the whole config files instead of the essential information


hayssams (approved):
LGTM

PR #1316, opened 2025-04-28
Fix/retry rate limit exception

tiboun:
# FEAT: Bigquery merge optimization on partitionned table
Given a table partitionned with date and having data for every hours (in our case, hourly observed weather), the target table contains 161,57 GB.

When SL_PRUNE_PARTITION_ON_MERGE is set to False,

Ingestion of one hourly observed day costs 161,57 GB in bigquery because it scans the whole table.

When it is set to True, then ingestion is done in two steps:
- detecting incoming dates: costs 20 MB
- apply merge for those dates with a replayed date: 
    - incoming data: 765,98 MB
    - target data: 498 MB
    - total merge job cost: 1,22 GB

If target data, doesn't exists, then it would costs only incoming data.

Merge can be optimized because:
- bigquery optimize projected column
- bigquery support partition pruning
- we know incoming partitions data
- partition column is used in merge keys

Therefore, if partition column is in merge keys, only data in that partition are concerned for the merge process. We just have to pre-compute partition column in order to make the filter determinist and have the ability prune partitions.

SCD2 hasn't been optimized in this PR but may be optimized by replacing {{ tableFullName }} with a CTE where tableFullName is filtered on partition to prune existing data. However, it requires testing of different use cases to be sure about the behavior.

# FIX: Add rate limit exception retry during table deletion

Bug fix, parallel table deletion may encounters quota limit. In order to maximize temporary table deletion, retry is applied. If table is not deleted, then temporary tables will expire on their own.

# FIX: premature shutdown of parallel executions
When an error occured, the thread pool were shutdown, preventing other tasks to continue and therefore having a message stating that tasks are submitted after pool is shutdown. Therefore, the new implementation waits for all tasks to be processed before checking the whole results and shutdown the pool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,29 @@
 
 # Release notes
 
-# 1.4.1:
+# 1.5.0:
+__Improvement__:
+- add `SL_PRUNE_PARTITION_ON_MERGE` settings to prune partition on merge. Currently for bigquery only. Recommended to be set to true for load scenarios with merge.
+- add `partitionPruningKey` and `quotedPartitionPruningKey` variables in write strategy templates. They are only set when target table's partition column is one of the merge keys and feature is enabled. 
+
 __Bug fix__:
 - fix index out of bound exception when extracting table name from file name
+- add rate limit exception retry during table deletion
+- fix premature shutdown of parallel executions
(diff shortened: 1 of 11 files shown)


hayssams (approved):
LGTM

hayssams:
GR8 PR
Thanks!

PR #1502, opened 2025-11-22
fix: Unprivileged container image

baptistegh:
## Summary

This PR introduces support for running container images in an unprivileged mode. It removes the requirement for elevated permissions and improves overall security by isolating runtime operations from host privileges.

Details:

* Adds an unprivileged image execution path
* Create a new user `starlake` inside the container
* This user has `UID=1001` and `GID=0` as requested in the Openshift guidelines.
* Theses changes also insure that a random uid can run the binaries without issues if he belongs to the `GID=0` group

Why:
Improves security posture, reduces attack surface, and aligns with best-practice least-privilege principles.

**Status: WIP**

**Breaking change? No**

### Other changes

Small fixes in the docker scripts.

### Remaining Todos

* [ ] Ensure the docker compose still works fine (it may require to add `user: 0` on the containers needing privileges).

## Contributor checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,13 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
+# Allow arbitrary UID
+RUN chgrp -R 0 /app && chmod -R g=u /app && \
+    chmod 755 /app/starlake.sh && \
+    useradd starlake -u 1001 -g 0 -m -s /bin/bash
+
+USER 1001
+
 COPY starlake/bin /app/bin
 COPY starlake/starlake.sh /app/starlake.sh
 COPY starlake/versions.sh /app/versions.sh
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -12,9 +12,10 @@ GREEN='\033[0;32m'
 NC='\033[0m' # No Color
 
 function usage() {
-    echo "Usage: $0 (-e [environment]) (-b)"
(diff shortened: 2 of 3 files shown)


scripts/docker-build.sh
```diff
@@ -81,7 +73,7 @@ fi
 
 echo Preparing docker image for ${ENVIRONMENT}
 if [ "$BUILD" == "true" ]; then
-  ./scripts/docker-prepare.sh -e $ENVIRONMENT -p
+  ./scripts/docker-prepare.sh -e $ENVIRONMENT -b # TODO: -b or -p ?
```
fupelaqu, line 76: there should be 2 specific entries, one for publishing and another one for building

Dockerfile
```diff
@@ -29,6 +29,13 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
+# Allow arbitrary UID
+RUN chgrp -R 0 /app && chmod -R g=u /app && \
```
fupelaqu, line 33: It will be necessary to update the docker-compose files in order to mount the different file systems (probably by adding an initialization phase for the affected containers with the necessary privileges to mount the file systems).

fupelaqu (commented):
LGTM except the impact on docker-compose

scripts/docker-build.sh
baptistegh: There was no `[ "$PUBLISH" = "true" ]` in this script.
The `-p` was given if `BUILD=true`.
Should I fix this ?

PR #1667, opened 2026-07-24
fix(schema): allow domains to share a landing directory when table patterns are disjoint

### LINKED ISSUE #1665: Allow multiple domains to share a landing directory when their table patterns are disjoint (managed file-receiver pattern)
## Summary

`SchemaHandler.checkDomainsVars` rejects two domains sharing the same `metadata.directory`:

```
ERROR SchemaHandler: Domain directory - <dir> is defined 2 times. A directory can only appear once in a domain definition file.
```

This makes a **shared landing zone** impossible to model, although it is a common integration pattern: an upstream file-transfer system (managed file receiver) drops every file of several functional domains into ONE directory, and the per-table filename patterns are the routing key.

## Use case

Migration of a legacy GCS ingestion: the file receiver writes all files to `gs://<filereceiver-bucket>/CRM2/Download` — tables belong to two different BigQuery datasets, hence two Starlake domains (`pack__crm2__client`, `pack__crm2__transactions`). All table `pattern`s are mutually disjoint, so routing is unambiguous.

Pointing both domains' `directory` at the landing zone is required for `preload --strategy imported` to see the incoming files (it lists the DOMAIN-level directory and filters on the tables' patterns — per-table `directory` overrides are not consulted by preload). But the uniqueness check flags it, and worse:

```scala
val directoryErrors = Utils.duplicates(...) match {
  case Right(_) =>
    if (!raw) {
      this._domains = Some(nonEmptyDomains)
      DatasetArea.initDomains(storage, nonEmptyDomains.map(_.name))   // SKIPPED on duplicates
    }
    Nil
  case Left(errors) => errors
}
```

on duplicates, `_domains` is not cached and **`DatasetArea.initDomains` never runs** — the pending/ingesting/archive areas are not initialized on a fresh environment, even though the rest of the pipeline (preload, stage, load, transform) works fine with the shared directory (verified empirically: preload correctly filters per-table patterns, a file matching another domain's table does not mark the domain ready).

## Proposal

Relax the domain-directory uniqueness check when the sharing domains' table patterns are mutually disjoint (the routing stays deterministic):

- if two domains share a `directory` AND their tables' `pattern`s do not overlap → accept (at most a WARNING);
- keep the ERROR when patterns overlap (genuinely ambiguous routing);
- in all accepted cases, still run `DatasetArea.initDomains` for every domain.

Alternative (or complement): have `preload` consult per-table `directory` overrides, which would let each domain keep a distinct nominal directory while tables point at the shared landing zone.

## Environment

- starlake core 1.5.16 (`SchemaHandler.scala` ~1036)

fupelaqu:
Fixes #1665

## Problem

`SchemaHandler.initDomains` rejected any two domains sharing the same `metadata.directory`, which makes a shared landing zone (managed file-receiver pattern) impossible to model. Worse, on duplicates `_domains` was not cached and `DatasetArea.initDomains` never ran, so the pending/ingesting/archive areas were left uninitialized on a fresh environment even though the rest of the pipeline handles a shared directory correctly.

## Change

- Replace the flat directory-duplicates check with `checkDomainDirectories`:
  - Domains may share a landing directory as long as no table filename pattern is declared verbatim by more than one table across the sharing domains (this also catches the same pattern declared twice inside one sharing domain).
  - Accepted sharing logs a WARNING reminding the user that only identical pattern texts are detected, and that patterns must be kept effectively disjoint for routing to stay deterministic.
  - A duplicated pattern in a shared directory raises an ERROR listing the directory, the sharing domains, the pattern and its declaring domains.
- In all accepted cases, `_domains` is cached and `DatasetArea.initDomains` runs for every domain.
- Caching/area-init is now also gated on domain-name uniqueness (`nameErrors`), preserving fail-fast when the same domain name is defined twice.

## Tests

New `SharedDomainDirectorySpec` (5 scenarios, all green locally):
- shared directory + disjoint patterns → no error, domains cached, stage/unresolved/archive/replay areas created for every domain;
- shared directory + same pattern in two domains → single "Domain directory" ERROR, nothing cached, no area created;
- same pattern declared by two tables of one sharing domain → ERROR;
- duplicated domain names sharing a directory (disjoint patterns) → name ERROR only, nothing cached;
- single domain per directory → behavior unchanged.

`SchemaInfo1HandlerSpec` and `SchemaInfo2HandlerSpec` also pass locally.

## Known limitations / follow-ups (surfaced by adversarial review)

- Overlap detection is textual (pattern-string equality). Regexes that differ as strings but match the same filenames (e.g. `clients-.*` vs `.*`) are accepted; the WARN log makes this explicit and effective disjointness remains the user's responsibility. True regex-intersection checking would need an automaton library.
- `listStageFiles` extracts and deletes `.zip/.gz/.tgz` archives found in the landing directory without pattern scoping; with a shared directory, one domain's stage run may consume a sibling domain's archives. Pre-existing behavior, newly reachable — worth a dedicated follow-up issue before recommending archives in shared landing zones.
- Directory identity is compared as a raw string (no normalization of trailing slash/`./`/case) — pre-existing, unchanged from the old check.
- No end-to-end routing test of the file-receiver scenario yet (preload/stage/load behavior over a shared directory was verified empirically in #1665).

### DIFF
--- a/src/main/scala/ai/starlake/schema/handlers/SchemaHandler.scala
+++ b/src/main/scala/ai/starlake/schema/handlers/SchemaHandler.scala
@@ -1033,20 +1033,10 @@ class SchemaHandler(storage: StorageHandler, cliEnv: Map[String, String] = Map.e
       case Left(errors) =>
         errors
     }
-    val directoryErrors = Utils.duplicates(
-      "Domain directory",
-      nonEmptyDomains.flatMap(_.resolveDirectoryOpt()),
-      s"%s is defined %d times. A directory can only appear once in a domain definition file."
-    ) match {
-      case Right(_) =>
-        if (!raw) {
-          this._domains = Some(nonEmptyDomains)
-          DatasetArea.initDomains(storage, nonEmptyDomains.map(_.name))
-
-        }
-        Nil
-      case Left(errors) =>
-        errors
+    val directoryErrors = checkDomainDirectories(nonEmptyDomains)
+    if (nameErrors.isEmpty && directoryErrors.isEmpty && !raw) {
+      this._domains = Some(nonEmptyDomains)
+      DatasetArea.initDomains(storage, nonEmptyDomains.map(_.name))
     }
 
     invalidDomainsFiles.foreach {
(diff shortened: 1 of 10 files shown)


hayssams (approved):
LGTM

PR #1668, opened 2026-07-24
fix(config): expose ConnectionInfo._transpileDialect in the application JSON schema

### LINKED ISSUE #1666: ConnectionInfo._transpileDialect cannot be set from application.sl.yml — JSON schema rejects it, making RUN-mode transpilation unreachable
## Summary

`ConnectionInfo._transpileDialect` exists in the model and is honored by the engine, but it **cannot be set from `application.sl.yml`**: the application JSON schema rejects it at load time.

```
ai.starlake.exceptions.SchemaValidationException: Invalid content for .../metadata/application.sl.yml:
     - $.application.connections.duckdb: property '_transpileDialect' is not evaluated and the schema does not allow unevaluated properties
```

## Context / why it matters

`SqlTransformations.transpileAndSubstituteSelectStatement` transpiles the main SELECT when:

```scala
if (test || connection._transpileDialect.isDefined) { ... SQLUtils.transpile(...) }
```

and `SQLUtils.transpilerDialect` resolves the source dialect from it:

```scala
conn._transpileDialect match {
  case Some(dialect) => JSQLTranspiler.Dialect.valueOf(dialect)
  ...
```

So a project whose transform SQL is written in BigQuery dialect can run its transforms locally on DuckDB (`transform` RUN mode) with the same BQ→DuckDB transpilation the unit-test harness applies — by declaring on the local connection:

```yaml
connections:
  duckdb:
    type: "jdbc"
    _transpileDialect: "GOOGLE_BIG_QUERY"
    options:
      url: "jdbc:duckdb:..."
```

But `starlake_AppConfigV1.json` only allows `type/sparkFormat/loader/quote/separator/options` on a connection, so loading fails before the engine ever sees the field. The feature is effectively unreachable from a YAML project.

## Repro

1. Add `_transpileDialect: "GOOGLE_BIG_QUERY"` to any connection in `metadata/application.sl.yml`.
2. Run any CLI command → `SchemaValidationException` above (validation of application.sl.yml is unconditional, `validateOnLoad` does not apply).

## Fix

Add `_transpileDialect` (string, enum of `JSQLTranspiler.Dialect` values) to the connection definition in the application JSON schema (`starlake_AppConfigV1.json` / ConnectionV1) — and to the doc: it is the supported way to run dialect-X SQL on a local engine.

## Environment

- starlake core 1.5.16

fupelaqu:
Fixes #1666

## Problem

`ConnectionInfo._transpileDialect` exists in the model and is honored by the engine (`SqlTransformations.transpileAndSubstituteSelectStatement`, `SQLUtils.transpilerDialect`), but it could not be set from `application.sl.yml`: the application JSON schema only allowed `type/sparkFormat/loader/quote/separator/options` on a connection, and strict validation (`unevaluatedProperties: false`) rejected the file before the engine ever saw the field. RUN-mode transpilation (e.g. running BigQuery-dialect transform SQL locally on DuckDB) was effectively unreachable from a YAML project.

## Change

- Add `_transpileDialect` to the `ConnectionV1` definition in both schema resources:
  - `starlake.json` — the aggregate schema actually loaded by `YamlSerde.validateConfigFile` at runtime;
  - `starlake_AppConfigV1.json` — the per-config fragment shipped to projects (IDE completion).
  The property is `["string", "null"]` with an enum of the `JSQLTranspiler.Dialect` values (`ANY`, `AMAZON_REDSHIFT`, `DATABRICKS`, `DUCK_DB`, `GOOGLE_BIG_QUERY`, `SNOWFLAKE`) plus `null`, so an empty placeholder value behaves like every sibling `ConvertibleToString` property. The enum gives IDE completion and catches typos at validation time. Jinja/env substitution runs before validation on the runtime path, so templated values keep working.
- `SQLUtils.transpilerDialect` now rejects an invalid declared dialect with an `IllegalArgumentException` listing the valid values, instead of a bare `Dialect.valueOf` stack trace — this covers config paths that bypass the JSON schema (typesafe `application.conf`, programmatic construction).

## Tests

New `TranspileDialectAppConfigSpec` (7 tests, green locally):
- a connection with `_transpileDialect` passes schema validation and the value survives in the node;
- end-to-end `Settings.loadApplication` → `AppConfig.connections(...)._transpileDialect` is populated (guards the pureconfig CamelCase mapping of the leading-underscore field);
- every `JSQLTranspiler.Dialect` value is accepted;
- an explicit null `_transpileDialect` is accepted and maps to `None`;
- an invalid value is rejected with a `SchemaValidationException` mentioning `_transpileDialect`;
- a sync test pins the enum in **both** JSON resources to `JSQLTranspiler.Dialect.values()` (duplicate-safe, sorted-list compare), so a transpiler upgrade adding a dialect fails the build until the schemas are updated;
- `SQLUtils.transpilerDialect` resolves a declared dialect and reports invalid ones with a meaningful error.

The `Arbitrary[ConnectionInfo]` generator in `YamlSerdeSpec` now also generates `_transpileDialect`, so the application round-trip property test exercises serialization/validation of the field.

## Notes (adversarial review outcomes)

- `Settings.loadApplication` validates content without a Jinja pass (unlike the runtime `loadApplicationYaml` path, which substitutes first). It has no caller in core main sources; external callers passing a Jinja-templated `_transpileDialect` through it would be rejected by the enum. Documented here as a known limitation of that entry point.
- The user-facing documentation for connections lives in the docs site repo; `_transpileDialect` should be documented there as the supported way to run dialect-X SQL on a local engine — docs follow-up.

### DIFF
--- a/src/main/resources/starlake.json
+++ b/src/main/resources/starlake.json
@@ -314,6 +314,19 @@
         "options": {
           "$ref": "#/definitions/MapString",
           "description": "Connection options"
+        },
+        "_transpileDialect": {
+          "type": ["string", "null"],
+          "enum": [
+            null,
+            "ANY",
+            "AMAZON_REDSHIFT",
+            "DATABRICKS",
+            "DUCK_DB",
+            "GOOGLE_BIG_QUERY",
+            "SNOWFLAKE"
+          ],
+          "description": "Source SQL dialect of the transform statements executed over this connection. When set, SELECT statements are transpiled from this dialect to the connection engine dialect at run time (e.g. run BigQuery-dialect SQL locally on DuckDB). ANY stands for generic SQL"
         }
       },
       "required": ["type"]
--- a/src/main/resources/starlake_AppConfigV1.json
+++ b/src/main/resources/starlake_AppConfigV1.json
@@ -464,6 +464,19 @@
         "options": {
           "$ref": "#/definitions/MapString",
           "description": "Connection options"
+        },
+        "_transpileDialect": {
+          "type": ["string", "null"],
+          "enum": [
+            null,
+            "ANY",
+            "AMAZON_REDSHIFT",
+            "DATABRICKS",
+            "DUCK_DB",
+            "GOOGLE_BIG_QUERY",
+            "SNOWFLAKE"
+          ],
+          "description": "Source SQL dialect of the transform statements executed over this connection. When set, SELECT statements are transpiled from this dialect to the connection engine dialect at run time (e.g. run BigQuery-dialect SQL locally on DuckDB). ANY stands for generic SQL"
         }
       },
       "required": [
--- a/src/main/scala/ai/starlake/sql/SQLUtils.scala
+++ b/src/main/scala/ai/starlake/sql/SQLUtils.scala
@@ -242,7 +242,12 @@ object SQLUtils extends LazyLogging {
 
   def transpilerDialect(conn: ConnectionInfo): JSQLTranspiler.Dialect =
     conn._transpileDialect match {
-      case Some(dialect) => JSQLTranspiler.Dialect.valueOf(dialect)
+      case Some(dialect) =>
+        Try(JSQLTranspiler.Dialect.valueOf(dialect)).getOrElse(
+          throw new IllegalArgumentException(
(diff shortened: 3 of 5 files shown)


hayssams (commented):
LGTM

PR #1696, opened 2026-08-14
Setup: install the API for real - move the right zip, overwrite on re-install, verify the swap

fupelaqu:
## Symptom (real Windows workstation)

`starlake serve` dies immediately with *"Le chemin d'accès spécifié est introuvable"* while the CLI itself works. `serve` does not start the CLI — it calls `bin\api\bin\local-run-api.bat`. On that machine `bin\api` contained **only `lib`**: `bin`, `ui` and `git` were gone. The install log had said:

```
Renaming C:\Users\...\starlake\bin\starlake-api-1.7.0 to C:\Users\...\starlake\bin\api
Failed to rename C:\Users\...\bin\api\starbake.zip to C:\Users\...\projects\demo\starbake.zip
Failed to rename C:\Users\...\bin\api\starbake.zip to C:\Users\...\projects\demo\tpch001.zip
```

Note the third line: it renames **starbake.zip** to **tpch001.zip**.

## Three defects in the same ~10 lines

**1. The tpch001 move uses the wrong source.**

```java
File tpch001Zip = new File(apiDir, "tpch001.zip");          // declared…
final File demoTpch001Zip = new File(demoDir, "tpch001.zip");
if (starbakeZip.exists() && !starbakeZip.renameTo(demoTpch001Zip)) {   // …never used
```

`tpch001Zip` is never referenced (javac reports it as unused). Two silent outcomes:

- **normal path**: the starbake move just succeeded, so `starbakeZip.exists()` is false, the guard short-circuits and **tpch001.zip is never installed**. Reproduced on a clean 1.7.1 install: `bin/api/tpch001.zip` still sits there (375 903 bytes) and `projects/demo/` has only `starbake.zip`.
- **when the starbake move fails first** (see #2): the guard passes and the installer writes starbake.zip's bytes into a file named `tpch001.zip`.

**2. `File.renameTo` cannot overwrite on Windows.** On any re-install `projects/demo/*.zip` already exists, so both moves fail and only print a line — that is the log above. Replaced with `Files.move(..., REPLACE_EXISTING)`, which overwrites and throws a real exception instead of returning a boolean.

**3. The api directory swap ignored both return values** while printing `Renaming <extracted> to <api>` as if it had worked:

```java
deleteRecursively(apiDir);      // File.delete() result dropped
...
extractedDir.renameTo(renamedDir);   // boolean dropped
```

On Windows a file another process holds open cannot be deleted. With a running API (`starlake serve`, or the server the VS Code extension starts) keeping `bin/api/lib/*.jar` open, the cleanup deletes `bin/`, `ui/` and `git/`, keeps `lib/`, the target directory survives — and the rename then silently does nothing, leaving the extracted tree behind as `bin/starlake-api-<version>`. That is exactly the reported state, and it is unrecoverable without manual cleanup.

`replaceDirectory()` now verifies the deletion and aborts with the actual remedy instead of claiming success:

```
Cannot replace <install>/bin/api: some of its files could not be deleted.
A running Starlake API (starlake serve, or the VS Code extension) holds them open -
stop it and run the installation again.
```

## Also

The pre-download `deleteRecursively(apiDir)` is gone: the previous installation now survives until the new tree is on disk, so a failed download no longer leaves the machine without an API.

`distrib/setup.jar` is rebuilt from this source (`javac --release 17`; manifest fields unchanged).

## Verification

Exercised on real files through the two helpers:

| Case | Result |
| --- | --- |
| both zips present | each destination gets **its own** content (`tpch001.zip` = tpch bytes, not starbake's); `bin/api` left empty |
| re-install (destinations already exist) | overwritten, no failure line |
| target directory not fully deletable (macOS `chflags uchg` — the portable stand-in for "file held open by another process" on Windows) | deletion leaves exactly `lib/` behind — **the reported signature** — and the installer prints the remedy and exits **1** instead of continuing (exit code read directly, not through a pipe) |
| jar smoke-run | `Please specify the target directory`, exit 1 |

Follows #1692 / #1693 / #1694 / #1695 in the same series: making the installation work, and fail honestly, on locked-down corporate machines.


hayssams (approved):
LGTM

PR #1723, opened 2026-08-28
fix(transform): OVERWRITE inserts by name, never by position

### LINKED ISSUE #1722: transform OVERWRITE never rebuilds the target schema: adding a column fails, reordering columns silently corrupts the data (all engines)
### Version

1.7.1 (reproduced with the released CLI; templates unchanged on `main` / 1.7.2-SNAPSHOT and identical at tag `v1.7.1`)

### Symptom

A transform whose target table already exists never rebuilds the table schema. Adding a column to the `.sql` and re-running fails:

```
starlake transform --name sales_analytics.customer_orders
Error: Binder Error: table customer_orders has 7 columns but 8 values were supplied
```

Worse, when the shape happens to stay compatible the run **succeeds and silently misaligns the data**. Reordering two same-typed columns in the SELECT (no column added, no schema change at all):

```sql
-- before: customer_id, customer_name, email, ...
-- after : customer_id, email, customer_name, ...
```

```
starlake transform --name sales_analytics.customer_orders   # exit 0, both expectations green
select customer_id, customer_name, email from sales_analytics.customer_orders limit 3;
1,john.doe@email.com,John Doe
2,jane.smith@email.com,Jane Smith
3,m.johnson@newdomain.com,Michael Johnson
```

Every email is now stored in `customer_name`. Nothing in the logs, the exit code, `audit.audit` or `audit.expectations` records that the table is wrong.

This is **not DuckDB-specific** — DuckDB only supplies the wording of the error message. See the root cause below.

### Reproduction

Any project with an `OVERWRITE` transform. Minimal loop:

1. `starlake transform --name <domain>.<task>` — first run, table created from the SELECT, green.
2. Add a column to the `.sql` (anywhere).
3. `starlake transform --name <domain>.<task>` — fails.

Observed on DuckDB with a 7-column target:

| change to the SELECT | result |
| --- | --- |
| column appended last | `Binder Error: table customer_orders has 7 columns but 8 values were supplied` |
| column appended last, after `--sync-apply` | works |
| column inserted mid-list, after `--sync-apply` | `Conversion Error: Could not convert string 'emily.brown@email.com' to INT64 when casting from source column email` |
| two same-typed columns swapped | **exit 0, data silently written to the wrong columns** |

### Root cause

`OVERWRITE` on an existing table is `TRUNCATE` + a **positional** `INSERT`. `src/main/resources/templates/write-strategies/duckdb/overwrite.j2`:

```jinja
TRUNCATE TABLE {{ tableFullName }};
INSERT INTO {{ tableFullName }} {{ selectStatement }}
```

Only `create.j2` (first run, table absent) derives the schema from the SELECT — see `TransformStrategiesBuilder.buildTransform`, which picks `CREATE` only when `targetTableExists` is false.

All eleven engine folders have the same shape, so every engine is affected, in one of two ways:

- **positional INSERT** — `duckdb`, `jdbc`, `postgresql`, `snowflake`, `spark`, `redshift`: column-count mismatch, or silent misalignment when the arity and types happen to line up;
- **explicit column list** — `bigquery`, `mysql`, `sqlserver`, `trino`: the list is `tableColumnsCsv`, which is `TableComponents.columnNames`, i.e. the columns resolved *from the SELECT* — so a new column produces a "column not found in table" error instead.

BigQuery transforms go through the same builder (`BigQueryAutoTask.runBQ` → `buildAllSQLQueriesMerged`), so they are not spared.

Two aggravating details:

1. **The schema-sync path cannot fix it.** `AutoTask.buildAllSQLQueries` emits `ALTER TABLE` statements only when `taskDesc.readyForSync()` — i.e. the task declares typed `attributes` in its `.sl.yml` — and it computes them from the **YAML attributes**, appending new columns at the **end** of the table. Since the INSERT is positional, this only works when the new column also happens to be last in the SELECT. A column added mid-SELECT still fails after `--sync-apply`, now with a cast error. Note that the SELECT's own column names are already resolved a few lines above, as `resolvedColumnNames`, so the information needed to align the INSERT is available at that point.

2. **A failed run leaves the table half-migrated.** On DuckDB the ALTERs run with autocommit on (`JdbcAutoTask.runJDBC`, `conn.setAutoCommit(sinkConnection.isDuckDb())` before the `Alters` block) while the main SQL rolls back. After the mid-SELECT failure above the target had `first_name` added at position 9 and the previous 13 rows intact — and every subsequent run fails identically, because the table's column order can no longer match the SELECT's.

### Suggested fix

Any of these would close the correctness hole; the last one is the cheapest:

- For `OVERWRITE`, use the engine's atomic replace where it exists (`CREATE OR REPLACE TABLE … AS <select>` on DuckDB, Snowflake, BigQuery), which is also what `AutoTask`'s own scaladoc already claims the strategy does — the doc comment above `buildAllSQLQueries` says `OVERWRITE: CREATE OR REPLACE TABLE ... AS SELECT`, and has not matched the templates for a long time.
- Otherwise make the INSERT **name-based rather than positional** on every engine (`INSERT INTO t (<resolved columns>) SELECT …`), and derive the ALTERs from `resolvedColumnNames` instead of the YAML attributes, so that adding a column anywhere in the SELECT works without a `.sl.yml` edit.
- At minimum, fail loudly: compare `resolvedColumnNames` against the target's actual columns before running, and refuse the run with a clear message. The silent-misalignment case is a data-corruption bug and should never reach the warehouse.

Whatever the resolution, the ALTER statements should share the main SQL's transaction so a failed run does not leave a half-migrated table.

hayssams:
Fixes #1722.

## Root cause, verified

`OVERWRITE` on an existing table renders `TRUNCATE TABLE ...; INSERT INTO <table> <select>` on duckdb, jdbc, postgresql, snowflake, spark (both branches), and redshift: a **positional** insert. `create.j2` (schema derived from the SELECT) only runs when the target table does not exist (`TransformStrategiesBuilder.buildTransform`). So after the first run, any divergence between the SELECT's column order and the table's either errors (count/cast) or, when arity and types happen to line up, **silently writes every value into the wrong column**.

Reproduced with a failing test before fixing (`DuckDbOverwriteSchemaSpec`): a same-arity, same-typed swap (`AS t(name, id, amount)` against a table `(id, name, amount)`) ran green and left `("Xavier", "10", 1000)` in the table, names in the `id` column.

## Fix

Every overwrite template now passes the explicit target column list, `{{ tableColumnsCsv }}` (the SELECT-resolved, engine-quoted names). This is the exact shape the **append templates already use on every engine**, and the overwrite templates on bigquery/mysql/sqlserver/trino already had, so it aligns overwrite with the established pattern rather than inventing one:

- duckdb, jdbc, postgresql, snowflake, spark: `INSERT INTO t(cols) <select>`
- redshift: `INSERT INTO t(cols) SELECT cols FROM #SL_INCOMING`

Behavior after the fix: reordered SELECT columns land by name; a column the table does not have fails loudly ("column X does not exist") instead of corrupting; adding a column at any position works with `--sync-apply` since the ALTER appends at the end and the insert no longer cares about position.

The `AutoTask` scaladoc claiming `OVERWRITE: CREATE OR REPLACE TABLE ... AS SELECT` is corrected to describe the real strategy.

## Verification

- New `DuckDbOverwriteSchemaSpec` watched failing first (silent misalignment reproduced), passes after the fix
- `sbt "testOnly ai.starlake.job.transform.* ai.starlake.job.strategies.*"`: 10 suites, 79 tests, all green

## Deliberately not in this PR (still open on the issue)

- Running the schema-sync ALTERs inside the main transaction: the `conn.setAutoCommit(sinkConnection.isDuckDb())` special case predates this change (commit "Insert by name") and likely works around a DuckDB DDL-in-transaction limitation; changing transaction semantics needs its own investigation
- Deriving sync ALTERs from `resolvedColumnNames` instead of YAML attributes (would make add-a-column work without a `.sl.yml` edit)
- The atomic `CREATE OR REPLACE TABLE` variant: rejected here because it drops grants, constraints, and partitioning that `TRUNCATE + INSERT` preserves; that trade-off deserves an explicit decision

https://claude.ai/code/session_01MWLuy2yBXT6ynzrrzapdjr

### DIFF
--- a/src/main/resources/reference-general.conf
+++ b/src/main/resources/reference-general.conf
@@ -298,7 +298,7 @@ maxInteractiveRecords = ${?SL_MAX_INTERACTIVE_RECORDS}
 
 
 syncSqlWithYaml = false
-#syncSqlWithYaml = ${?SL_SYNC_SQL_WITH_YAML}
+syncSqlWithYaml = ${?SL_SYNC_SQL_WITH_YAML}
 
 syncYamlWithDb = true
 syncYamlWithDb = ${?SL_SYNC_YAML_WITH_DB}
--- a/src/main/resources/templates/write-strategies/duckdb/overwrite.j2
+++ b/src/main/resources/templates/write-strategies/duckdb/overwrite.j2
@@ -3,5 +3,5 @@ DROP MATERIALIZED VIEW IF EXISTS {{ tableFullName }};
 CREATE MATERIALIZED VIEW {{ tableFullName }} AS {{ selectStatement }}
 {% else %}
 TRUNCATE TABLE {{ tableFullName }};
-INSERT INTO {{ tableFullName }} {{ selectStatement }}
+INSERT INTO {{ tableFullName }}({{ tableColumnsCsv }}) {{ selectStatement }}
 {% endif %}
--- a/src/main/resources/templates/write-strategies/jdbc/overwrite.j2
+++ b/src/main/resources/templates/write-strategies/jdbc/overwrite.j2
@@ -3,5 +3,5 @@ DROP MATERIALIZED VIEW IF EXISTS {{ tableFullName }};
 CREATE MATERIALIZED VIEW {{ tableFullName }} AS {{ selectStatement }}
 {% else %}
 TRUNCATE TABLE {{ tableFullName }};
-INSERT INTO {{ tableFullName }} {{ selectStatement }}
+INSERT INTO {{ tableFullName }}({{ tableColumnsCsv }}) {{ selectStatement }}
 {% endif %}
--- a/src/main/resources/templates/write-strategies/postgresql/overwrite.j2
+++ b/src/main/resources/templates/write-strategies/postgresql/overwrite.j2
@@ -3,5 +3,5 @@ DROP MATERIALIZED VIEW IF EXISTS {{ tableFullName }};
 CREATE MATERIALIZED VIEW {{ tableFullName }} AS {{ selectStatement }}
 {% else %}
 TRUNCATE TABLE {{ tableFullName }};
-INSERT INTO {{ tableFullName }} {{ selectStatement }}
+INSERT INTO {{ tableFullName }}({{ tableColumnsCsv }}) {{ selectStatement }}
 {% endif %}
--- a/src/main/resources/templates/write-strategies/redshift/overwrite.j2
+++ b/src/main/resources/templates/write-strategies/redshift/overwrite.j2
@@ -6,7 +6,7 @@ DROP MATERIALIZED VIEW IF EXISTS {{ tableFullName }};
 CREATE MATERIALIZED VIEW {{ tableFullName }} AS SELECT {{ tableColumnsCsv }} FROM #SL_INCOMING;
 {% else %}
 TRUNCATE TABLE {{ tableFullName }};
-INSERT INTO {{ tableFullName }} SELECT {{ tableColumnsCsv }} FROM #SL_INCOMING;
+INSERT INTO {{ tableFullName }}({{ tableColumnsCsv }}) SELECT {{ tableColumnsCsv }} FROM #SL_INCOMING;
 {% endif %}
 
 DROP VIEW IF EXISTS #SL_INCOMING;
--- a/src/main/resources/templates/write-strategies/snowflake/overwrite.j2
+++ b/src/main/resources/templates/write-strategies/snowflake/overwrite.j2
@@ -5,5 +5,5 @@ DROP MATERIALIZED VIEW IF EXISTS {{ tableFullName }};
 CREATE MATERIALIZED VIEW {{ tableFullName }} AS {{ selectStatement }}
 {% else %}
 TRUNCATE TABLE {{ tableFullName }};
-INSERT INTO {{ tableFullName }} {{ selectStatement }}
+INSERT INTO {{ tableFullName }}({{ tableColumnsCsv }}) {{ selectStatement }}
 {% endif %}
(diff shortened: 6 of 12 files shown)


hayssams:
Added on top of the fix: when a run fails because the SELECT returns columns the target table does not have, the error now names the missing column(s) and prints the exact fix instead of the raw engine message:

```
Transform 'mydb.mytable' failed: the SELECT returns column(s) 'email' that table mydb.mytable does not have.
To add the missing column(s) to the table, sync the task attributes from your SQL, then re-run the transform:
  starlake transform --name mydb.mytable --sync-apply
  starlake transform --name mydb.mytable
Alternatively, declare the new column(s) in the attributes section of the task's .sl.yml.
Underlying error: ...Binder Error: Table "mytable" does not have a column with name "email"
```

The diagnosis runs only after a failure, never throws itself, and passes unrelated failures through untouched. Wired into both JdbcAutoTask (WHERE 1=0 probe) and BigQueryAutoTask (getBQSchema); BigQuery could not be integration-tested here (needs a live account), so the shared logic is covered by the DuckDB test and the BQ wiring is compile-verified.

hayssams:
syncSqlWithYaml is now real (was dead config: default false, env-var line commented out, only code reference inside a commented block). With SL_SYNC_SQL_WITH_YAML=true, a user-invoked transform aligns the task YAML attributes with the SQL before running, so a column added anywhere in the SELECT flows into the schema-sync ALTERs and the name-based INSERT with no manual step. Guards: user-invoked transforms only (audit/metrics/expectations/ingestion paths pass syncSchema=false), never audit tables, and a sync failure falls back to declared attributes with a warning instead of failing the run. Verified end to end on DuckDB: SQL-only task, new mid-SELECT column, run succeeds, values land by name, and the .sl.yml gains the synced attributes.

The failure hint now also mentions the env var as the automatic alternative.

Default deliberately stays false: enabling it means transform runs rewrite .sl.yml files in the project, which is a behavior change to make explicitly, not silently. Flip the default in reference-general.conf if you want it on out of the box.

hayssams:
Content landed on master via the branch-1.8 merge (62404e64b): the five commits were cherry-picked to branch-1.8 in #1724, which was then merged into master, so this PR's diff is now empty. Also on branch-1.7 via #1725. Closing as superseded; all changes are live on all three lines.

PR #1790, opened 2026-09-14
feat(run): run log, resumable runs and the run store backends

hayssams:
## What

P3 of `starlake run`: every run writes an event log, and a failed run can be continued instead of restarted.

```
starlake run --select +sales
  ... run id 01JF2K…, agg.daily FAILED

starlake run --resume            # or --resume-id 01JF2K…
  ... dom1.things SKIPPED (succeeded in attempt 1)
  ... agg.daily STARTED
```

## The pieces

**Event model** (`RunLog.scala`) — runs are an append-only event stream; `RunHistory.fold` replays every attempt into current per-task state. Nothing mutates a recorded row.

**`RunStore` seam** (`RunStore.scala`) — the trait speaks only in runs and events: no paths, no line offsets, no SQL. `RunStoreContractSpec` is shared by every backend and is written to fail an in-memory-only implementation, so a store that forgets to persist cannot pass it.

Three implementations:
- `FileRunStore` — the zero-config default, JSONL under `<root>/.starlake/runs`.
- `JdbcRunStore` — PostgreSQL, the schema owned by starlake-api (`run-store/postgres-schema.sql`). Selected by `SL_RUN_STORE_URL`.
- `NoopRunStore` — the off switch for `--dry-run` and `--no-run-log`. Not a third backend: reading is empty and reopening fails, because resuming a disabled log is a usage error, not a run with no history.

**Fingerprint** (`RunFingerprint.scala`) — graph, selection, options and env hashed separately so a refused resume names *which* input changed rather than saying "the project changed". Task bodies are digested per task and compared selectively: transforms contribute their SQL, load tables contribute nothing, since a load node's behaviour is driven by whatever files are in the stage area. A fingerprint part recorded before it existed counts as changed, not as equal.

**CLI** (`RunCmd`, `RunConfig`) — `--resume`, `--resume-id <id>`, `--force`, `--no-run-log`, plus `SL_RUN_LOG=false`. `--select`, `--exclude` and `--options` are rejected alongside a resume: both are replayed from the recorded run. Exit 4 = resume refused because the project changed.

## Environment

| Variable | Effect |
|---|---|
| `SL_RUN_STORE_URL` / `_USER` / `_PASSWORD` | Use the JDBC store. Blank is treated as absent |
| `SL_RUN_CORRELATION_ID` | The caller's own job id, stored beside the run (the runner keeps the run id) |
| `SL_RUN_LOG_DIR` | Where the file store writes |
| `SL_RUN_LOG` | `false`/`0`/`no`/`off` disables the log for every run |

A project rooted on remote storage with no `SL_RUN_LOG_DIR` falls back to a temp directory namespaced by a hash of the root, so two remotely-rooted projects on one host cannot read each other's runs, and the user is told once.

## Testing

`sbt Test/compile "testOnly ai.starlake.job.run.*"` — 199 tests, 16 suites, 0 failures, including `JdbcRunStoreSpec` against a PostgreSQL testcontainer. `scalafmtCheck` and `Test/scalafmtCheck` clean.

Note: `scalafmtSbtCheck` fails on `project/Versions.scala` and `project/PythonLibs.scala`. That is pre-existing drift on master, untouched by this branch.

https://claude.ai/code/session_01QDzMrrHcbajSQsMQNtrkTW


hayssams:
Suspending `starlake run` rather than extending it: a project is already run from the CLI with `starlake load` and `starlake transform`, so the runner has to earn its place before the run log and resume are worth landing.

Nothing here is lost. The complete feature, this P3 stack included, is preserved on `feature/starlake-run-suspended`. The removal of the runner from master is #1791.

PR #1793, opened 2026-09-18
fix(bigquery): terminate every presql statement in the native script

### LINKED ISSUE #1792: presql is glued to the main statement on the native BigQuery path (missing separator)
## Summary

On the **native BigQuery path**, any task carrying a non-empty `presql` builds a single
script in which the last `presql` statement is glued to the main statement, with no
separator. BigQuery then rejects the whole job:

```
Syntax error: Expected end of input but got keyword CREATE at [1:141]
```

The task never runs, and no amount of metadata can work around it — the statement
terminator is deliberately removed earlier in the pipeline.

## Root cause — three links

1. **`SchemaHandler.scala:1242`** (and `:1388`) — every `presql` is normalised at load time:

   ```scala
   presql = task.presql.map(sql => SQLUtils.stripComments(sql)),
   ```

2. **`SqlCommentStripper.scala:95-96`** — the last act of `stripComments` is to drop the
   statement terminator:

   ```scala
   // remove last ';' if any
   if (res.trim.endsWith(";")) res.trim.dropRight(1) else res
   ```

3. **`BigQueryAutoTask.scala:501`** — the native branch (`loadedDF = None`) concatenates
   without re-adding one:

   ```scala
   preSql.mkString(";\n") + mainSql() + ";\n" + postSql.mkString(";\n")
   ```

   `mkString(";\n")` separates the entries **from each other** but appends nothing after
   the last one, and `mainSql()` strips its own trailing `;` (line 317, *"Because if we
   have two ';' BQ fails"*).

So a single `presql` entry yields `DELETE ...` immediately followed by
`CREATE TABLE ...`.

`postsql` is unaffected: `mainSql()` is already followed by `";\n"`.

## Why it is not seen everywhere

Only the native path concatenates. When a Spark DataFrame is loaded
(`loadedDF = Some(df)`), `BigQueryAutoTask` calls `runPrePostSql(preSql)` and executes the
statements as separate jobs (lines 340 and 407), which works. The JDBC path does the same
via `runSqls` (`JdbcAutoTask.scala:268`), so DuckDB unit tests never exercise the defect —
they stay green with a broken configuration.

## Reproduction

Any transform task on a BigQuery connection **without** `_transpileDialect`, e.g.:

```yaml
version: 1
task:
  table: "my_table"
  domain: "my_domain"
  writeStrategy:
    type: "APPEND"
  presql:
    - "DELETE FROM my_domain.my_table WHERE d >= CURRENT_DATE();"
```

```
starlake transform --name my_domain.my_table
```

Adding or removing the trailing `;` in the YAML makes no difference — step 2 removes it
either way. Both the folded (`>`) and the quoted single-line form behave identically.

## Affected versions

Present, byte-identical and at the same line numbers, on:

- `branch-1.7` (1.7.4-SNAPSHOT, tip 2026-09-08) — and in the released 1.7.2 / 1.7.3
- `master` (1.8.8-SNAPSHOT, tip 2026-09-18)

## Suggested fix

Give every `presql` statement its own terminator, which also keeps the empty-list case
correct (`mkString` with a prefix would emit a stray leading `;`):

```scala
val allSql =
  preSql.map(_ + ";\n").mkString + mainSql() + ";\n" + postSql.mkString(";\n")
```

Draft PRs follow for both branches.

fupelaqu:
Fixes #1792.

On the native BigQuery path (`loadedDF = None`), `BigQueryAutoTask` built the script as

```scala
preSql.mkString(";\n") + mainSql() + ";\n" + postSql.mkString(";\n")
```

`mkString` separates the entries **from each other** but appends nothing after the last
one, so the final `presql` statement was glued to the main `CREATE`/`INSERT`:

```
Syntax error: Expected end of input but got keyword CREATE at [1:141]
```

A trailing `;` in the YAML cannot work around it: `SchemaHandler.scala:1242` runs every
`presql` through `SQLUtils.stripComments` at load time, and that function's last act is
`// remove last ';' if any` (`SqlCommentStripper.scala:95-96`). `mainSql()` strips its own
for the same reason (line 317).

```diff
-preSql.mkString(";\n") + mainSql() + ";\n" + postSql.mkString(";\n")
+preSql.map(_ + ";\n").mkString + mainSql() + ";\n" + postSql.mkString(";\n")
```

Giving each statement its own terminator also keeps the empty-list case correct — a
`mkString` with a prefix would emit a stray leading `;` for a task with no `presql`.
`postsql` needs nothing: `mainSql()` is already followed by `";\n"`.

Only the native path concatenates. The Spark path (`runPrePostSql`, lines 340 and 407) and
the JDBC path (`runSqls`, `JdbcAutoTask.scala:268`) execute the statements separately,
which is why DuckDB unit tests stay green on a configuration that is broken in production.

**Draft**: opened as a draft because the change has not been exercised by an automated
test yet. It was found against a real BigQuery run — the failing job's query text shows the
`DELETE` and the `CREATE TABLE` adjacent on the same line. Happy to add a unit test on the
script-assembly if you point me at the right place for it.

A companion PR targets `branch-1.7`, where the defect is byte-identical.


fupelaqu:
Build check: `sbt clean compile` succeeds on this branch under Java 17 (Zulu 17.0.11), with zero `[error]` lines, and `scalafmt` leaves the changed line as written — 98 columns, under the configured `maxColumn = 100`.

Still a draft only because no automated test exercises the script assembly. Glad to add one if you point me at the right place.

The `branch-1.7` backport is #1794; its CI also needed #1757 (the Slack workflow removal) cherry-picked, which that PR carries.

fupelaqu:
The red `test-only` run on this PR is pre-existing on `master`, not caused by this change.

**No test fails.** 23 suites *abort* during initialisation, all on the same error:

```
java.lang.NoSuchMethodError: 'void org.conscrypt.Conscrypt.setNamedGroups(javax.net.ssl.SSLSocket, java.lang.String[])'
  at com.google.api.gax.httpjson.HttpJsonConscryptUtils.lambda$configureConscryptSecurityProvider$0(HttpJsonConscryptUtils.java:112)
```

`google-api-gax` calls a Conscrypt method that is not in the conscrypt version on the
classpath, so every HTTPS call to a Google API fails — which takes out all the BigQuery
integration suites.

Same shape on the `workflow_dispatch` run of `test-only` against `master` itself
(run 34779224296, 2026-09-13, before this PR existed):

| | master, 2026-09-13 | this PR |
| --- | --- | --- |
| Suites aborted | 23 | 23 |
| Tests failed | 0 | 0 |
| Tests succeeded | 878 | 861 |

Every `test-only` run listed for the last two weeks is red the same way, on unrelated
branches. Looks like a dependency-alignment issue worth its own issue — happy to open one
if it is not already tracked.

hayssams (approved):

PR #1794, opened 2026-09-18
fix(bigquery): terminate every presql statement in the native script (1.7 backport)

fupelaqu:
Backport of #1793 to `branch-1.7`. Refs #1792.

The defect is byte-identical on this branch — same code, same line numbers:

- `SchemaHandler.scala:1242` / `:1388` — every `presql` goes through `SQLUtils.stripComments`
- `SqlCommentStripper.scala:95-96` — `// remove last ';' if any`
- `BigQueryAutoTask.scala:501` — `preSql.mkString(";\n") + mainSql()`, nothing after the last entry

```diff
-preSql.mkString(";\n") + mainSql() + ";\n" + postSql.mkString(";\n")
+preSql.map(_ + ";\n").mkString + mainSql() + ";\n" + postSql.mkString(";\n")
```

Worth backporting because 1.7 is what is deployed: on the native BigQuery path, **any**
task carrying a non-empty `presql` fails at run time with

```
Syntax error: Expected end of input but got keyword CREATE at [1:141]
```

and no metadata change can work around it, since the terminator is removed at load time.

**Draft**, same reason as #1793: no automated test exercises the script assembly yet.


fupelaqu:
## CI on this branch was red for an unrelated reason — fixed by backporting #1757

The two failing `deploy` jobs came from `.github/workflows/slack.yml`, not from the change
in this PR. The job log shows the webhook resolving to an empty string:

```
env:
  SLACK_WEBHOOK_URL:
##[error]Specify secrets.SLACK_WEBHOOK_URL
```

`secrets.SLACK_STARLAKE_CORE_WEBHOOK_URL` no longer exists, so `8398a7/action-slack@v3`
aborts on every event. `master` does not hit this because the workflow was removed there in
`5761d5d21` — *"ci: drop the Slack notifications, the team moved to Discord (#1757)"* — and
`branch-1.7` never received it.

I cherry-picked that commit as-is (`-x`, original authorship preserved). It applied cleanly
and removes the same five things it removed on `master`:

- `.github/workflows/slack.yml` (deleted)
- the Slack step in `snapshot-docker-no-tests.yml`, `snapshot-docker-with-tests.yml`,
  `snapshot-jar-only.yml` and `test-only.yml`

No Slack reference is left on the branch. The other differences between these workflows and
their `master` counterparts are pre-existing branch divergences (`pull_request_target`,
`paths-ignore`, the pyspark pin) and are untouched.

Happy to split this into its own PR if you would rather keep the backport separate from the
BigQuery fix — it is a self-contained commit.

---

The BigQuery change itself compiles: `sbt compile` succeeds on this branch under Java 17,
and `scalafmt` leaves the line as written (98 columns, under the configured 100).

